PR #519, opened 2026-04-14
fix(bedrock): strip empty arrays and prevent false-positive model cooldowns

Bowl42:
## Summary
- **Sanitizer**: Strip empty `system:[]` and `tools:[]` arrays that Bedrock rejects but Anthropic API tolerates
- **Error classification**: Replace broad `contains("model")` check with `isBedrockModelUnavailable()` using specific Bedrock error patterns — prevents field validation errors from being misclassified as `ScopeModel` which would wrongly freeze the provider
- **Error fixer**: Add empty array stripping as reactive fallback for requests that bypass the proactive sanitizer (e.g. raw/relay mode)

## Root cause
Bedrock ValidationException errors containing the word "model" (e.g. field validation) were being misclassified as model-unavailability errors, triggering provider cooldowns. Additionally, empty arrays sent by clients caused preventable 400 errors.

## Test plan
- [x] 2 tests for empty array stripping (empty vs non-empty preserved)
- [x] 7 tests for `isBedrockModelUnavailable` (4 true positives, 3 false positive rejections)
- [x] All existing adapter tests pass

## 发布说明

* **Bug 修复**
  * 改进了 Bedrock 模型不可用错误的检测和分类，现在能更准确地识别特定的模型相关错误场景。
  * 修复了请求处理中的空字段问题，确保空的顶级 `system` 和 `tools` 字段被正确删除。
  * 优化了 HTTP 400 响应的错误处理逻辑，提供更精确的错误诊断。

### DIFF
--- a/internal/adapter/provider/bedrock/adapter.go
+++ b/internal/adapter/provider/bedrock/adapter.go
@@ -379,8 +379,10 @@ func classifyBedrockHTTPError(statusCode int, body []byte, headers http.Header, 
 	case statusCode == 400:
 		proxyErr.Scope = domain.ScopeRequest
 		proxyErr.Retryable = false
-		if strings.Contains(bodyLower, "validationexception") && model != "" &&
-			(strings.Contains(bodyLower, "model") || strings.Contains(bodyLower, "inference profile")) {
+		// Only upgrade to ScopeModel for genuine model-availability errors.
+		// Use specific Bedrock error patterns to avoid false positives from
+		// field validation errors that happen to mention "model".
+		if model != "" && isBedrockModelUnavailable(bodyLower) {
 			proxyErr.Scope = domain.ScopeModel
 			proxyErr.Reason = domain.CooldownReasonModelUnavailable
 			proxyErr.Model = model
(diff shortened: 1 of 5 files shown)


Bowl42 (commented):
## Claude Review ✅

**Sanitizer (`sanitizer.go`)**: Empty array stripping is clean — only removes `system` and `tools` when they're empty arrays (`v.IsArray() && v.Get("#").Int() == 0`). Non-empty arrays and missing fields are untouched.

**Error classification (`adapter.go`)**: `isBedrockModelUnavailable()` uses specific Bedrock error patterns instead of the overly broad `contains("model")`. The 6 patterns cover real model-access errors (foundation model resolution, authorization, access denied, inference profile). Field validation errors that happen to mention "model" will correctly stay as `ScopeRequest` — no false cooldowns.

**Error fixer (`000_bedrock.go`)**: Empty array stripping added as step 4, matching the sanitizer logic. This covers requests that bypass the proactive sanitizer (raw mode, relay).

**Case normalization refactor**: Good call from CodeRabbit — `isBedrockModelUnavailable` now handles its own `ToLower`, reducing caller coupling.

No concerns. LGTM.

awsl233777 (approved):
Checks are green, approving for merge.

PR #530, opened 2026-04-23
fix(bedrock): resolve bare anthropic.<short> via discovery

Bowl42:
## Summary
- When a client sends a bare foundation-model ID like `anthropic.claude-sonnet-4-6`, Bedrock rejects the invoke for models that have no on-demand foundation SKU (`on-demand throughput isn't supported — retry with an inference profile`). `resolveModelID` previously passed these through unchanged, so the call always failed even when discovery had an invoke-ready profile under the short name.
- Strip the `anthropic.` prefix and consult the discoverer when the ID is bare (not region-prefixed, not dated+versioned). On a hit, return the discovered invoke-ready ID; on a miss, fall back to the existing passthrough so explicit client intent still wins when discovery is unavailable.

## Context
Observed as the second failure in a chain: a request whose primary route returned `Invalid signature in thinking block` failed over to a Bedrock route carrying the bare foundation ID, which Bedrock then refused with the on-demand message. This PR only addresses the Bedrock ID resolution side of that chain; the signature handling is a separate concern.

## Test plan
- [x] `go test ./internal/adapter/provider/bedrock/...` passes, including two new cases:
  - `client-supplied bare anthropic.X resolves via discovery`
  - `client-supplied bare anthropic.X with discovery miss falls back to passthrough`
- [ ] Live Bedrock smoke: send `anthropic.claude-sonnet-4-6` with discovery enabled and confirm the request reaches an inference profile rather than returning the on-demand error.

## 发布说明

* **Bug Fixes**
  * 优化了模型ID解析逻辑，增强了对Anthropic模型的智能识别和自动转换功能。

* **Tests**
  * 添加了新的测试用例以验证模型ID解析的准确性和容错能力。

### DIFF
--- a/internal/adapter/provider/bedrock/model_mapping.go
+++ b/internal/adapter/provider/bedrock/model_mapping.go
@@ -2,6 +2,7 @@ package bedrock
 
 import (
 	"regexp"
+	"strings"
 )
 
 // modelDatePattern matches an Anthropic short name that already carries an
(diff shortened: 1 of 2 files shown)


Bowl42:
@coderabbitai review

awsl233777 (approved):
Checks are green, and the Bedrock bare anthropic.<short> discovery fallback looks correct.

PR #532, opened 2026-04-23
fix(custom): strip thinking blocks when upstream rejects signature

Bowl42:
## Summary
- Signatures on \`thinking\` / \`redacted_thinking\` content blocks are scoped to the deployment that produced them (account, model, and sometimes a private wrapper format). When a client replays extended-thinking history across providers, the destination rejects the request with bodies like:
  > \`400 {\"error\":{\"message\":\"messages.0.content.0: Invalid \\\`signature\\\` in \\\`thinking\\\` block ...\"}}\`
- New fixer matches the exact backticked phrase on a 400 from a Claude-format client, strips every \`thinking\` / \`redacted_thinking\` block from \`messages[].content[]\`, and lets the retry layer re-send. Plain-string content is left untouched; a message that becomes all-thinking ends up with an empty content array rather than being dropped (keeps turn ordering intact).

## Context
Observed as the 422/400 failure path when a custom-proxy route forwards extended-thinking history to an upstream that can't verify the incoming signatures. The match is tight enough to skip AWS SigV4 signature errors and thinking-budget validation errors, which both share individual keywords but not the backticked phrase.

## Test plan
- [x] \`go test ./internal/adapter/provider/custom/...\` passes, including:
  - \`TestThinkingSignatureFixer_MatchResponse\` (positive + 5 negative cases)
  - \`TestStripThinkingBlocks\` (mixed content, string content, all-thinking message, no-messages payload)
- [ ] Live smoke: replay a session with extended-thinking history against a custom-proxy upstream that rejects cross-source signatures; confirm the retry succeeds.

## 发布说明

* **Bug Fixes**
  * 改进了对 Claude API 特定签名验证错误的处理，增强了请求重试机制的可靠性

* **Tests**
  * 添加了相应的测试用例以确保错误处理的准确性和稳定性

### DIFF
--- a/internal/adapter/provider/custom/error_fixer/100_thinking_signature.go
+++ b/internal/adapter/provider/custom/error_fixer/100_thinking_signature.go
@@ -54,13 +54,20 @@ func (f *thinkingSignatureFixer) FixRequest(req *http.Request, body []byte) (*ht
 	return req, stripThinkingBlocks(body)
 }
 
+// emptyAssistantPlaceholder is the content we leave on an assistant
+// message that was nothing but thinking blocks. Upstreams reject both
+// empty content arrays and zero-length text, and dropping the whole
+// message would merge adjacent user turns (also rejected). A minimal
+// text block keeps turn structure intact for the retry.
+var emptyAssistantPlaceholder = []byte(`[{"type":"text","text":"[thinking omitted]"}]`)
+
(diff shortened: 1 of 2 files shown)


Bowl42:
@coderabbitai review

awsl233777 (approved):
Checks are green and the fix direction is acceptable. Approving for merge.

PR #533, opened 2026-04-23
feat(bedrock): discover 4.6/4.7, persist catalog, fix adaptive thinking

Bowl42:
## Summary

Fixes a cluster of Bedrock-side regressions exposed when AWS shipped Claude Opus 4.6 / 4.7 and Sonnet 4.6:

- **Discovery found none of the new profiles.** AWS serves them under `type=APPLICATION` (not `SYSTEM_DEFINED`) with bare short-name IDs (`us.anthropic.claude-opus-4-7`, `us.anthropic.claude-opus-4-6-v1`). Dropped the type filter; unified `profileIDPattern` + `foundationModelPattern` into one `bedrockAnthropicPattern` that accepts region-prefixed dateless shapes.
- **FM-only 4.x entries were being indexed** as direct-invoke targets and then failing at request time with "on-demand throughput isn't supported". Skip entries whose `inferenceTypesSupported` lacks `ON_DEMAND`.
- **Graceful degrade** when a model is entitled but its profile hasn't landed yet: `claude-opus-4-7` → `4-6` → `4-5` → `4`, logging a WARN so operators see the substitution.
- **Persist the catalog in SQLite** (`bedrock_discovery_entries`, keyed per-provider). Restart no longer pays the ~1–5s `ListInferenceProfiles` cold-start. Admin refresh button now POSTs to force-refresh (bypasses rate-limit); page loads stay cheap GETs.
- **Opus 4.7 rejects classic thinking** (`thinking.type=enabled`). Added `adaptThinkingForModel`: for the adaptive-only set (currently just `claude-opus-4-7`), rewrite to `thinking.type=adaptive` + `output_config.effort` with a budget→effort mapping so classic-shape clients (Claude Code CLI) can still hit it.
- Layering follows existing repo convention (`domain.BedrockDiscoveryEntry` + `repository.BedrockDiscoveryRepository` interface + sqlite impl, one-line wire-up in `cmd/maxx`, no shims).

## Test plan

- [x] `go test ./...` green
- [x] `MAXX_BEDROCK_LIVE=1` live tests pass (discovery list + direct invoke on 4-5/4-6/4-7)
- [x] Docker Claude Code CLI → maxx → Bedrock end-to-end: `claude-opus-4-7`, `claude-opus-4-6`, `claude-sonnet-4-6`, `claude-sonnet-4-5` all return real model output
- [x] Restart of maxx: GET `/bedrock-models` served from DB in ~5ms; POST forces a ~4s AWS re-fetch
[...]

PR #534, opened 2026-04-23
fix(bedrock): expose bedrock-models under self-service /api/providers

Bowl42:
## Summary

After #533 merged, `GET /api/providers/{id}/bedrock-models` on production (maxx.gemini.skr.rocks) returned **404**. Root cause: the handler was only registered under `/api/admin/providers/{id}/bedrock-models`, but the frontend's default axios baseURL is `/api`, so every non-admin page-load hit the self-service path — which had no route for `bedrock-models`.

Fix:
- Add the 4-segment `providers/{id}/bedrock-models` case to `self_service.go`.
- Extract the shared GET/POST logic into `serveBedrockDiscoveredModels` so both handlers call the same code path (no duplication, no drift risk).

## Test plan

- [x] `go test ./...` green
- [x] Verified locally: with a real Bedrock provider, both `/api/admin/providers/{id}/bedrock-models` and `/api/providers/{id}/bedrock-models` return 200 OK + the same 14-model catalog; POST force-refresh returns 200 on both.

## 发行说明

* **新功能**
  * 为自助服务 API 添加了 Bedrock 模型端点。
  * 仅管理员用户可访问和执行模型发现刷新操作。

* **改进**
  * 优化了已发现模型的缓存机制，减少不必要的查询。
  * 管理员界面中的刷新功能现已受权限控制。

### DIFF
--- a/cmd/maxx/main.go
+++ b/cmd/maxx/main.go
@@ -16,7 +16,7 @@ import (
 	"time"
 
 	"github.com/awsl-project/maxx/internal/adapter/client"
-	_ "github.com/awsl-project/maxx/internal/adapter/provider/bedrock" // Register bedrock adapter
+	"github.com/awsl-project/maxx/internal/adapter/provider/bedrock"
 	_ "github.com/awsl-project/maxx/internal/adapter/provider/claude"  // Register claude adapter
 	_ "github.com/awsl-project/maxx/internal/adapter/provider/custom"  // Register custom adapter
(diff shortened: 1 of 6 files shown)


Bowl42:
@coderabbitai full review

Bowl42:
@coderabbitai full review

Bowl42:
@coderabbitai full review

awsl233777 (approved):
Approved: PR is open, mergeable, non-draft, and all required checks are passing.

PR #548, opened 2026-04-30
fix(bedrock): recover from thinking-block envelope rejections

Bowl42:
## Summary
- Anthropic validators reject thinking-block envelopes produced by a different deployment with two phrases: `` Invalid `signature` in `thinking` block `` and `` Invalid `data` in `redacted_thinking` block ``. Common triggers are router switches, multi-account rotation, or clients restoring history into a different backend.
- Match the family with a single regex `` Invalid `[^`]+` in `(?:thinking|redacted_thinking)` block `` — generic over the field name (future-proofs new validations on the same blocks) but pinned on the block type so we never strip thinking blocks for an unrelated error (e.g. `tool_result`).
- Bedrock direct adapter does a one-shot retry after stripping; custom adapter routes through the existing `error_fixer` pipeline. Both share `bedrock.IsThinkingBlockEnvelopeError` and `bedrock.StripThinkingBlocks`.
- Stripping removes both `thinking` and `redacted_thinking` together so a request carrying both doesn't pay a second retry.

## Test plan
- [x] `go build ./...`
- [x] `go test ./internal/adapter/provider/bedrock/... ./internal/adapter/provider/custom/...`
- [ ] Manual: replay a transcript captured against Anthropic against Bedrock and confirm the request now succeeds (one extra round-trip for the strip+retry).

## Summary by CodeRabbit

* **New Features**
  * 适配器在遇到“思考块”类的400响应时会尝试一次自动重试：如果重试前后请求体被改写，重试成功则不会将首次被拒绝的400暴露为响应信息。

* **Bug Fixes**
  * 改进了对思考块相关验证错误的检测与分类，并在特定不可用情形下使发现缓存失效以避免重复故障。

* **Tests**
  * 增加了覆盖思考块检测、移除与重试行为的单元测试，包含多种正反案例验证。


Bowl42:
@coderabbitai full review please

awsl233777 (changes requested):
Blocking issue found.

The new `thinking_envelope` fixer does not actually run for a common custom-adapter Bedrock error shape. `FindFixers` is priority-exclusive: once a priority-0 fixer matches, priority-100 fixers are skipped. The existing `bedrockFixer` matches any 400 body containing `Bedrock Runtime` or `InvokeModel`, so a Bedrock SDK-style error such as:

```text
operation error Bedrock Runtime: InvokeModel ... ValidationException: messages.0.content.0: Invalid `data` in `redacted_thinking` block
```

selects only `bedrock`, not `thinking_envelope`. That retry strips generic Bedrock fields, leaves the thinking/redacted_thinking blocks intact, and the next round stops as “bedrock already applied”. This breaks the PR’s stated custom-adapter coverage for the redacted/thinking envelope case.

I verified this locally with a focused temporary test: `FindFixers(&http.Response{StatusCode: 400}, body, domain.ClientTypeClaude)` returns `bedrock`, not `thinking_envelope`, for a body containing both `Bedrock Runtime` and `Invalid \`data\` in \`redacted_thinking\` block`.

Suggested fix: make the thinking-envelope path win before the broad Bedrock fixer, e.g. either:

- give `thinkingEnvelopeFixer` higher priority than `bedrockFixer` and update the filename/tests accordingly, or
- have `bedrockFixer.MatchResponse` decline when `bedrock.IsThinkingBlockEnvelopeError(body)` is true, letting the narrow fixer handle it.

Please add a regression test for the combined Bedrock SDK + thinking-envelope error body so this does not regress.

Validation I ran:
- `go test ./internal/adapter/provider/bedrock/... ./internal/adapter/provider/custom/...` ✅
- `go test ./...` ❌ locally only because the top-level package cannot find embedded `web/dist` (`main.go:26:12: pattern all:web/dist: no matching files found`); package tests otherwise reached/pass through the touched packages.
- `go build ./...` ❌ same missing local `web/dist` embed asset.

Bowl42 (commented):
@awsl233777 thanks — confirmed the bug. Fixed in ec73782 by having `bedrockFixer.MatchResponse` decline when `bedrock.IsThinkingBlockEnvelopeError(body)` is true (your Option B), so the priority-100 `thinking_envelope` fixer wins for envelope rejections wrapped in Bedrock SDK chains.

Regression coverage:
- Registry-level: `TestFindFixers_BedrockDefersToThinkingEnvelope` (asserts `bedrock` not in returned fixers, `thinking_envelope` is) for the exact body shape you described.
- Matcher-level: two new cases in `TestBedrockFixer_MatchResponse` for `data`/`redacted_thinking` and `signature`/`thinking` wrapped in `Bedrock Runtime: InvokeModel` SDK chains.

Bowl42:
@coderabbitai full review please

awsl233777 (approved):
Re-reviewed the follow-up commits. The previously requested blocker is fixed: `bedrockFixer.MatchResponse` now defers when `bedrock.IsThinkingBlockEnvelopeError(body)` matches, so the narrow `thinking_envelope` fixer can handle Bedrock SDK-wrapped thinking/redacted_thinking envelope errors. Regression coverage was added at the registry level and matcher level for both `data`/`redacted_thinking` and `signature`/`thinking` variants.

Validation I ran locally:
- `go test ./internal/adapter/provider/bedrock/... ./internal/adapter/provider/custom/...` ✅

Note: I did not rerun top-level `go test ./...` / `go build ./...` locally because this checkout lacks `web/dist` for the embed in `main.go`; GitHub CI is the authoritative gate for that path.

awsl233777:
@Bowl42 Current merge gate is still blocked only by the `CodeRabbit` status being `PENDING` since 2026-04-30 11:40 UTC. The earlier code-review blocker has been fixed and re-approved; GitHub CI is green.

Please wait for CodeRabbit to finish, or retrigger/investigate it if it stays stuck. Once that status turns green, this PR should be merge-ready from the visible gates.

PR #549, opened 2026-05-03
feat(retention): split request-detail retention into success/failed

Bowl42:
## Summary
- Adds opt-in toggle `request_detail_retention_split_enabled` so successful requests can be cleaned faster while failed ones (FAILED/CANCELLED/REJECTED) are kept longer for debugging.
- New settings `request_detail_retention_seconds_success` / `_failed` fall back to the existing unified key — flipping the switch alone is a no-op until at least one side is set explicitly, so upgrades preserve current behavior.
- `ClearDetailOlderThan` (request + attempt repos) takes an optional status filter; background cleanup runs two passes under split mode and uses `min(success, failed)` of positive values as the dynamic interval.
- Executor immediate-clear (`retention=0`) keeps its current semantics for split=false; under split=true it only fires when both sides are `0`, because the request status isn't known at ingress.
- Settings UI gains a "Configure success/failed separately" switch that swaps the unified input for two split inputs.

## Test plan
- [x] `go test ./internal/core/... ./internal/repository/sqlite/... ./internal/executor/...`
- [x] `go test ./tests/e2e -run RequestDetailRetention`
- [x] `npm run typecheck` + `npm run build` in `web/`
- [ ] Manually toggle the switch in Settings and verify each input persists / falls back as expected

## Notes
- New integration tests cover: status-filter SQL on both repos, dev_mode shielding, the JOIN-by-parent-status path for attempts, and four split-mode scenarios in `task_test.go` (split off, split on with success-only clearing, fallback to unified, and `-1` no-op).
- Tests backdate `created_at` instead of injecting a fake clock — flagged as a known follow-up if a clock abstraction is needed later.

## Summary by CodeRabbit

* **新功能**
  * 新增请求详情保留策略的“分离模式”：可通过开关分别为成功/失败请求设置独立保留时长，UI 切换后显示独立输入并隐藏原单一配置项。

* **文档**
  * 更新系统设置文档，新增分离保留模式及回退行为说明。

* **国际化**
  * 添加中/英本地化文案以支持新 UI 字段与说明。

* **测试**
  * 增强端到端与仓库层测试，覆盖分离模式下的各类保留场景。


Bowl42:
@coderabbitai full review please

awsl233777 (approved):
Reviewed the split request-detail retention change locally.

No blocking issues found. The core contracts look safe:
- split=false preserves unified retention behavior for immediate clear decisions;
- split=true defers status-dependent cleanup until final request status is known;
- PENDING/IN_PROGRESS are intentionally excluded from split buckets to avoid clearing in-flight request bodies;
- request/attempt repository filters preserve dev_mode shielding and use parent request status for attempts.

Validation run:
- `go test ./internal/core/... ./internal/repository/sqlite/... ./internal/executor/...` ✅
- `go test ./tests/e2e -run RequestDetailRetention` ✅
- `cd web && pnpm install` ✅
- `cd web && pnpm run typecheck` ✅
- `cd web && pnpm run build` ✅
- `cd web && pnpm run lint` ✅

Note: I tried to follow the PR body literally for frontend npm validation, but this frontend uses `pnpm@10.7.0` and only has `pnpm-lock.yaml`; `npm ci` fails because there is no npm lockfile. I verified using the actual CI/package-manager path instead.

LGTM.

Bowl42:
@coderabbitai full review please

PR #554, opened 2026-05-08
fix: harden codex responses stream conversion

awsldev:
# Summary
- Forward native OpenAI Responses/Codex SSE events (`response.*`) when the OpenAI→Codex response converter receives an already-Responses-shaped stream.
- Preserve the terminal `data: [DONE]` sentinel once and persist duplicate suppression even when `[DONE]` is received without earlier convertible chunks.
- Preserve upstream error semantics (`event: error`, `type: error`, or non-chat error payloads) instead of silently swallowing errors into ambiguous timeouts.
- Support event-only native Responses SSE frames such as `event: response.completed` even when the JSON body lacks `type`.
- Synthesize `response.completed` before `[DONE]` for started Chat Completions streams that end without a `finish_reason`, so Codex clients still see the completion contract.
- Format multiline SSE `data:` payloads with a `data:` prefix on every line.
- Keep the existing Chat Completions chunk → Responses event synthesis path intact.

# Why
Codex clients wait for `response.completed` before treating the SSE stream as complete. The previous converter handled Chat Completions chunks, but if an upstream/provider already emitted Responses-style events, those events did not have `choices[]` and were dropped. That could swallow `response.completed`, causing Codex-side reconnects or upstream timeout/502 symptoms even when the upstream stream itself had completed.

# Research notes
- Checked similar public transformer patterns for Responses stream handling. The relevant event contract is the same set handled/forwarded here: `response.output_text.delta`, `response.output_item.added`, `response.function_call_arguments.delta`, and `response.completed`.
- Rechecked local Codex provider tests/adapter behavior: the provider explicitly treats a stream closed before `response.completed` as an error, so preserving or synthesizing that event before terminal stream end is required.
- Re-reviewed failure/compatibility edges: error events keep error semantics, event-only native Responses frames are accepted, `[DONE]` is de-duplicated, and multiline SSE data is emitted in valid SSE framing.

# Tests
- `go test ./internal/converter` — pass
- `go test ./internal/converter ./internal/adapter/provider/codex ./internal/executor/responsemodifier ./tests/e2e` — pass
- `go test ./...` — pass

### DIFF
--- a/internal/converter/coverage_misc_sse_test.go
+++ b/internal/converter/coverage_misc_sse_test.go
@@ -42,3 +42,26 @@ func TestSSE_FormatStringData(t *testing.T) {
 		t.Fatalf("expected string data")
 	}
 }
+
+func TestSSE_FormatMultilineDataPrefixesEveryLine(t *testing.T) {
+	out := FormatSSE("response.completed", []byte("{\n  \"type\": \"response.completed\"\n}"))
+	outStr := string(out)
+	for _, want := range []string{
+		"event: response.completed",
+		"data: {",
+		"data:   \"type\": \"response.completed\"",
+		"data: }",
+	} {
+		if !strings.Contains(outStr, want) {
+			t.Fatalf("expected %q in %q", want, outStr)
+		}
+	}
+
+	events, remaining := ParseSSE(outStr)
+	if remaining != "" || len(events) != 1 {
+		t.Fatalf("expected one complete parsed event, remaining=%q events=%d", remaining, len(events))
+	}
+	if events[0].Event != "response.completed" || !strings.Contains(string(events[0].Data), "response.completed") {
+		t.Fatalf("unexpected parsed event: %+v", events[0])
+	}
+}
--- a/internal/converter/coverage_openai_stream_test.go
+++ b/internal/converter/coverage_openai_stream_test.go
@@ -216,6 +216,54 @@ func TestOpenAIToCodexStreamForwardsDoneWhenSplitFromCompletion(t *testing.T) {
 	}
 }
 
+func TestOpenAIToCodexStreamForwardsNativeResponsesEvents(t *testing.T) {
+	state := NewTransformState()
+	respConv := &openaiToCodexResponse{}
+
+	created := map[string]interface{}{
+		"type": "response.created",
+		"response": map[string]interface{}{
+			"id":         "resp_native",
+			"object":     "response",
+			"created_at": int64(1),
+			"status":     "in_progress",
+		},
+	}
+	delta := map[string]interface{}{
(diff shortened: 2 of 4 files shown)


awsldev:
Updated after investigating the Codex reconnect / 502 timeout path.

Root cause addressed:
- OpenAI→Codex stream conversion now forwards native Responses/Codex `response.*` SSE events instead of dropping them when they are not Chat Completions chunks.
- This preserves `response.completed`, which Codex provider/client code requires before accepting the stream as complete.
- Existing `[DONE]` forwarding still remains de-duplicated.

Validation run locally:
- `go test ./internal/converter`
- `go test ./internal/converter ./internal/adapter/provider/codex ./internal/executor/responsemodifier ./tests/e2e`
- `go test ./...`

Latest commit: `fbc1c15`.

awsldev:
@coderabbitai re-review

awsldev:
Second hardening pass pushed: `ef96fae`.

Additional edge cases covered/fixed:
- Native Responses SSE event names are normalized from JSON `data.type`, so a mismatched `event:` header cannot hide `response.completed` from clients that care about event names.
- Upstream SSE error events (`event: error`, `type: error`, or non-chat error payloads) are forwarded instead of being silently swallowed and later surfacing as an ambiguous timeout.

Validation repeated locally after this pass:
- `go test ./internal/converter`
- `go test ./internal/converter ./internal/adapter/provider/codex ./internal/executor/responsemodifier ./tests/e2e`
- `go test ./...`

awsldev:
@coderabbitai re-review

awsldev:
Third hardening pass pushed: `d8b631c`.

Additional issues fixed after conservative review:
- `event: error` now wins over conflicting `data.type=response.*`, preserving upstream error semantics.
- Event-only native Responses frames such as `event: response.completed` are forwarded even when the body lacks `type`.
- Started Chat Completions streams that end with `[DONE]` but no `finish_reason` now synthesize terminal item/content done events plus `response.completed` before forwarding `[DONE]`.
- Multiline SSE payloads now emit `data:` on every line instead of producing invalid strict SSE framing.

Validation repeated locally after this pass:
- `go test ./internal/converter`
- `go test ./internal/converter ./internal/adapter/provider/codex ./internal/executor/responsemodifier ./tests/e2e`
- `go test ./...`

awsldev:
@coderabbitai re-review

awsldev:
Closing this stale PR. The branch is now conflicting with main, and this streaming/converter hardening work is too old to be safely revived in-place. If any of these Codex/OpenAI responses stream hardening behaviors are still needed, they should be revalidated against current main and reintroduced as a new, smaller PR instead of continuing on this historical branch.

PR #555, opened 2026-05-11
fix(bedrock): strip temperature/top_p/top_k when thinking is on

Bowl42:
## Summary
- Bedrock rejects `temperature` / `top_p` / `top_k` whenever extended thinking is active. For always-on adaptive SKUs (e.g. Opus 4.7) this happens even without an explicit `thinking` block, so requests like `{model: claude-opus-4-7, temperature: 0.7}` 400'd.
- Three-layer defense rather than a model-list patch:
  1. **Rule-driven** — `SanitizeForBedrockCompat` strips the three params when `thinking.type` is `enabled` or `adaptive`.
  2. **Model-specific** — `adaptThinkingForModel` calls `StripSamplingParams` for adaptive-only models so always-on adaptive is covered.
  3. **Error-driven retry** — adapter retries once on Bedrock 400 matching `IsSamplingParamRejectedError`, mirroring the existing thinking-envelope retry. Future SKUs / message wording shifts recover automatically without a code change.

## Test plan
- [x] `go test ./internal/adapter/... -count=1`
- [x] `go build ./...`
- [x] New unit tests: `TestSanitizeForBedrockCompatStripsSamplingParamsWhenThinking`, `TestIsSamplingParamRejectedError`, `TestStripSamplingParams`, `TestAdaptThinkingForModelStripsSamplingParams`
- [ ] Manual: replay the failing request (`claude-opus-4-7` + `temperature: 0.7`) against Bedrock and confirm 200

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## 发布说明

* **新功能**
  * 增加模型短名归一化和思维策略适配，针对特定自适应模型自动调整或移除采样参数以确保兼容性。

* **Bug修复**
  * 在 400 响应路径新增针对采样参数被拒的独立重试逻辑：若检测到此类错误，则尝试移除采样参数并重试一次，提升请求成功率和鲁棒性。

* **测试**
  * 添加多项单元测试，覆盖采样参数清理、错误识别和思维策略适配行为。

[Review Change Stack](https://app.coderabbit.ai/change-stack/awsl-project/maxx/pull/555)

PR #558, opened 2026-05-13
fix(bedrock): catch "temperature is deprecated" wording for retry

Bowl42:
## Summary
- Bedrock changed its sampling-param rejection wording for \`claude-opus-4-7\`. Production logs since 2026-05-12 show 87+ requests failing with \`\`\`\`\`temperature\`\`\` is deprecated for this model.\`\`\`\` — no mention of \"thinking\".
- The PR #555 regex was tightened to require a thinking-state phrase (to avoid false-positives on \`temperature must be between 0 and 1; thinking budget exceeds max_tokens\`), which made the new deprecation wording slip through.
- Add a separate alternation that triggers on \`is deprecated\` in proximity to a sampling field name. Same recovery (strip + replay).

## Why this is safe
Unlike the bare-co-occurrence case, \"X is deprecated for this model\" near a sampling field name is unambiguous — it can only be the case we want to recover from. Added a guard test for an unrelated \"this endpoint is deprecated\" message to verify the regex doesn't over-match.

## Test plan
- [x] New cases in \`TestIsSamplingParamRejectedError\`: backticks, no-backticks, reversed clause, unrelated-deprecation guard
- [x] All existing cases still pass
- [ ] CI green

## Summary by CodeRabbit

* **Bug Fixes**
  * 改进了系统对来自服务端的采样参数相关错误消息的识别，新增对多种“已弃用”格式的识别规则，提升错误分类的准确性与稳健性。

* **Tests**
  * 扩展了测试用例，覆盖新增的弃用消息变体并加入防误报场景，增强回归保护与可靠性验证。

[Review Change Stack](https://app.coderabbit.ai/change-stack/awsl-project/maxx/pull/558)

### DIFF
--- a/internal/adapter/provider/bedrock/sanitizer.go
+++ b/internal/adapter/provider/bedrock/sanitizer.go
@@ -162,27 +162,36 @@ func StripSamplingParams(body []byte) []byte {
 // samplingParamRejectedPattern matches Bedrock 400 validation messages that
 // reject `temperature` / `top_p` / `top_k` because the target model is in
 // (possibly always-on) extended-thinking mode. Wording shifts across
-// releases — the Anthropic-style format wraps the field in backticks
-// ("`temperature` may only be set to 1 when thinking is enabled") while
-// the Bedrock-native format drops them — and the field may appear before
-// or after the thinking clause.
+// releases:
 //
-// The pattern requires both (a) one of the rejected field names and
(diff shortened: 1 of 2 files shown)


awsl233777 (approved):
Approved after all required checks passed, local Bedrock sanitizer focused test passed, and the stuck Playwright required check was rerun successfully.

PR #565, opened 2026-05-17
fix(retention): unblock detail cleanup on large tables

Bowl42:
## Summary

Detail cleanup (`ClearDetailOlderThan`) was monopolizing the SQLite single-writer lock on million-row tables, starving normal API writes (线上 "卡死"). Fix has three parts:

### 1. Yield writer lock between batches
- Batch size 500 → 200 (smaller write transactions).
- 50ms sleep between batches.
- Drain-to-completion within a single invocation (no per-call batch cap — earlier cap caused regressions because the cursor reset each tick).

### 2. Index that actually gets used
Added migration v13 creating `idx_proxy_requests_detail_cleanup` and `idx_proxy_upstream_attempts_detail_cleanup`:
- **SQLite**: partial index `(created_at, id) WHERE (request_info IS NOT NULL OR response_info IS NOT NULL) [AND dev_mode = 0]` — rows leave the index when their details are nulled, so steady-state scans touch only eligible rows.
- **MySQL**: composite `(created_at, id)` (no partial index syntax).

Cleanup queries rewritten so the planner actually uses the index:
- Cursor switched from `id > ?` to `(created_at, id) > (lastCreatedAt, lastID)` and `ORDER BY (created_at, id)`. Without this, SQLite falls back to the PK btree and ignores the partial index.
- For attempt cleanup, parent filter switched from `IN (subquery)` to `EXISTS correlated subquery`. The IN form drove from the parent table via `idx_proxy_upstream_attempts_proxy_request_id` plus a TEMP B-TREE sort.

Two new tests (`TestProxyRequestClearDetailOlderThan_UsesPartialIndex`, `TestProxyUpstreamAttemptClearDetailOlderThan_UsesPartialIndex`) run `EXPLAIN QUERY PLAN` and assert the cleanup query hits the partial index — regression guard against accidentally undoing the cursor/EXISTS shape.

### 3. Safe migration on huge tables ⚠️ operator action required
SQLite has no online `CREATE INDEX`; on a multi-million-row table the migration would block writes for tens of seconds (same class of issue the PR is fixing, just shifted to startup). To avoid that:

- `runDetailCleanupIndexMigration` checks `COUNT(*)`; tables ≤ 500k rows build the index inline (sub-second on SSD).
- Tables > 500k rows: **migration v13 is marked applied but the index is NOT built**. A clear log entry prints the exact SQL to run manually during a maintenance window.

When the index is missing, cleanup still works correctly — it falls back to a full scan, but the batch + sleep yielding from part 1 keeps writer-lock contention bounded, so 卡死 is resolved at the algorithm level regardless of index presence. Large-installation operators can apply the index at their convenience for the full performance benefit.

Log lines to watch for after upgrade on large installs:
```
[Migration v13] SKIPPING idx_proxy_requests_detail_cleanup build: proxy_requests has N rows ...
  CREATE INDEX IF NOT EXISTS idx_proxy_requests_detail_cleanup ON proxy_requests(...) WHERE ...
```

## Test plan

- [x] `go test ./internal/repository/sqlite/ ./internal/core/ ./tests/e2e/` — all pass
- [x] `EXPLAIN QUERY PLAN` confirms `idx_proxy_requests_detail_cleanup` and `idx_proxy_upstream_attempts_detail_cleanup` are both used by their respective cleanup queries (covered by the two new tests)
- [x] Existing `TestProxyRequestClearDetailOlderThan` multi-batch / dev-mode / status-filter cases still pass with the new tuple cursor
- [ ] On a production-sized DB:
  - confirm 卡死 is resolved (API writes are no longer starved by cleanup)
  - if `proxy_requests` > 500k rows, apply the printed manual CREATE INDEX during a maintenance window for the full speedup

### DIFF
--- a/internal/repository/sqlite/migrations.go
+++ b/internal/repository/sqlite/migrations.go
@@ -375,33 +375,30 @@ var migrations = []Migration{
 	},
 	{
 		Version:     13,
-		Description: "Add created_at indexes on proxy_requests and proxy_upstream_attempts for detail-cleanup",
+		Description: "Add detail-cleanup indexes on proxy_requests and proxy_upstream_attempts",
 		Up: func(db *gorm.DB) error {
-			// detail 清理 (ClearDetailOlderThan) 用 `WHERE created_at < ?` 过滤过期行；
-			// 在百万级以上的请求日志表上，没有 created_at 索引会导致每次清理都全表扫描，
-			// 进而长时间占用 SQLite 单写锁，表现为线上 API 写入"卡死"。
-			switch db.Dialector.Name() {
-			case "mysql":
-				if err := db.Exec("CREATE INDEX idx_proxy_requests_created_at ON proxy_requests(created_at)").Error; err != nil && !isMySQLDuplicateIndexError(err) {
-					return err
-				}
-				if err := db.Exec("CREATE INDEX idx_proxy_upstream_attempts_created_at ON proxy_upstream_attempts(created_at)").Error; err != nil && !isMySQLDuplicateIndexError(err) {
-					return err
-				}
-				return nil
-			default:
-				if err := db.Exec("CREATE INDEX IF NOT EXISTS idx_proxy_requests_created_at ON proxy_requests(created_at)").Error; err != nil {
-					return err
-				}
-				return db.Exec("CREATE INDEX IF NOT EXISTS idx_proxy_upstream_attempts_created_at ON proxy_upstream_attempts(created_at)").Error
-			}
+			// detail 清理 (ClearDetailOlderThan) 用
+			//   WHERE created_at < ? AND (request_info IS NOT NULL OR response_info IS NOT NULL) [AND dev_mode = 0]
+			//   ORDER BY id LIMIT ?
+			// 之前没索引时每次都全表扫，长时间占 SQLite 单写锁导致线上写入"卡死"。
+			//
+			// SQLite 用 **partial index**：WHERE 子句让只有"还没清理"的行进入索引；行被 null
+			// 后自动从索引移除，存量收敛后索引很小，scan 几乎免费。键设计 (created_at, id) 同
+			// 时服务 cleanup 的 ORDER BY id LIMIT。
+			//
+			// MySQL 不支持 partial index（functional index 语法又不同），退化为普通的
+			// (created_at, id) 复合索引。
+			//
+			// 注意：大表（>1M 行）首次 CREATE INDEX 会一次性扫描全表并阻塞写入数秒到几十秒，
+			// 这是一次性成本（之后清理快得多）。下面 log 行让运维能看到这段窗口。
+			return runDetailCleanupIndexMigration(db)
 		},
 		Down: func(db *gorm.DB) error {
 			switch db.Dialector.Name() {
 			case "mysql":
 				for _, sql := range []string{
-					"DROP INDEX idx_proxy_requests_created_at ON proxy_requests",
-					"DROP INDEX idx_proxy_upstream_attempts_created_at ON proxy_upstream_attempts",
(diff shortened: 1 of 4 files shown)


awsldev:
Blocking concerns before merge:

1. **PR description/test plan is stale after the later commits.**
   The body still says migration v13 creates `idx_proxy_requests_created_at` / `idx_proxy_upstream_attempts_created_at` on startup and that each cleanup invocation is capped at 200 batches. The current patch has moved to `idx_*_detail_cleanup` and removed `maxBatchesPerCall`, so the PR description and test plan no longer match the actual code.

2. **The large-table migration behavior needs to be explicit in the PR body/release notes.**
   `runDetailCleanupIndexMigration` now marks v13 as applied while skipping index creation above `detailCleanupIndexRowThreshold`. That is a reasonable safety tradeoff for SQLite writer-lock risk, but operators need to know that large installations must run the printed manual SQL during a maintenance window; otherwise they may assume the cleanup index exists after upgrade when it does not.

3. **Please include query-plan validation for the final index shape.**
   The cleanup query combines `created_at < ?`, detail-field predicates, `id > ?`, and `ORDER BY id LIMIT ?`. Before merging a performance fix for production-sized tables, please confirm with `EXPLAIN QUERY PLAN` (or equivalent) that the final partial/composite index is actually used as intended on SQLite, especially after the “skip index on large table” fallback is introduced.

Non-blocking nit: CodeRabbit’s duplicated cleanup constants suggestion is tidy-up only; I would not block merge on that by itself.

Bowl42:
@awsl233777 thanks — all three concerns addressed. Walking through each:

**1. PR description stale** — rewrote the PR body to match the actual code (new index names `idx_*_detail_cleanup`, `maxBatchesPerCall` cap removed, current batch + cursor shape documented).

**2. Large-table migration behavior** — now explicit in the PR description under "Safe migration on huge tables ⚠️ operator action required". v13 prints a clear log entry with the exact CREATE INDEX SQL when row count > 500k, and the body documents that operators must apply it manually. Cleanup remains correct without the index — algorithm-side batch+sleep yields the writer lock either way; index only adds steady-state speedup.

**3. EXPLAIN QUERY PLAN validation** — the validation surfaced two real issues that I fixed before re-pushing:

   - The original `id > ?` cursor + `ORDER BY id` made SQLite fall back to the PK btree, never picking the partial index. Switched cursor to `(created_at, id)` tuple with matching `ORDER BY (created_at, id)`.
   - The attempt cleanup with `proxy_request_id IN (subquery)` drove from the parent table and added a `USE TEMP B-TREE FOR ORDER BY`. Switched to `EXISTS` correlated subquery.

   Final plans (verified):

   \`\`\`
   proxy_requests cleanup:
     SEARCH proxy_requests USING INDEX idx_proxy_requests_detail_cleanup (created_at<?)

   proxy_upstream_attempts cleanup:
     SEARCH a USING INDEX idx_proxy_upstream_attempts_detail_cleanup (created_at<?)
     CORRELATED SCALAR SUBQUERY 1
       SEARCH proxy_requests USING INTEGER PRIMARY KEY (rowid=?)
   \`\`\`

   Locked in by two new tests (\`TestProxyRequestClearDetailOlderThan_UsesPartialIndex\`, \`TestProxyUpstreamAttemptClearDetailOlderThan_UsesPartialIndex\`) that EXPLAIN the cleanup query and assert the expected index name appears in the plan. Anyone reverting the cursor shape or IN/EXISTS form will hit the test failure.

Pushed as commit \`45c1a65\`.

Bowl42:
@coderabbitai review

awsl233777 (approved):
Approved after maintainer blocker follow-up: PR description is current, large-table migration/operator action is documented, EXPLAIN QUERY PLAN validation is covered by regression tests, and required checks are green.

PR #561, opened 2026-05-15
feat: multi-instance coordination via Redis

Bowl42:
## Summary

Adds a Redis-backed coordinator so multiple `maxx` instances can share a
database and Redis without stepping on each other. The original
single-instance behavior is preserved when Redis is not configured.

The work is sequenced so each commit is independently reviewable; see
[docs/multi-instance-rfc.md](docs/multi-instance-rfc.md) for the design
boundaries.

## What's in the box

**Coordinator subsystem** — `internal/coordinator/`
- Three-mode config (`standalone` / `fail-fast` / `degraded`) selected by
  `MAXX_COORDINATOR_MODE`; `degraded` falls back to memory and logs WARN
  on background reconnect but does not hot-swap implementations
- Pub/Sub + KV + instance heartbeat primitives
- `coordinator.RedisClient(c)` helper exposes the underlying client to
  cooldown's store without polluting the Coordinator interface

**Stale request reclamation rewrite** — `internal/repository/sqlite/proxy_request.go`
- `MarkStaleAsFailed(aliveInstanceIDs []string)` replaces the
  `currentInstanceID string` signature
- 60s orphan grace + 30min hard timeout + empty-list safety door
- Live instances reclaim dead peers' orphans on a 45s periodic sweep

**Cooldown made distributed-safe** — `internal/cooldown/`
- New `CooldownStore` sub-interface (memory + Redis impls) holds the
  cross-instance truth; Lua-backed `SetIfLater` keeps "later wins"
  atomic under race
- Generation counter per provider drives reload; events carry only
  `{provider_id, generation}` so a lost pub/sub message only delays
  reconvergence, never permanently desyncs instances
- Every public method on `cooldown.Manager` keeps its existing signature
  (no `ctx`, no `error` cascades into callers)

**Cache invalidation across instances** — `internal/repository/cached/`
- 7 cached repos publish structured `{entity, op, id, updated_at}` events
- Subscribers v1 still do full-table reload; payload reserves shape for
  later targeted invalidation without bumping the wire protocol

**Session hot cache** — `internal/repository/cached/session.go`
- Local → Coordinator KV → DB read-through with KV backfill, 1h TTL
- DB stays the source of truth so a Redis outage at most adds a round-trip

**Unified wiring** — `internal/core/coordinator_setup.go`
- `SetupCoordinator` + `AttachCachedReposToCoordinator` used by both
  `cmd/maxx/main.go` and the desktop launcher; desktop passes
[...]

PR #566, opened 2026-05-17
perf(retention): leader-only cleanup, MySQL index reorder, faster batch

Bowl42:
## Summary

生产反馈驱动的三项优化（#565 后续）：

1. **Leader-only cleanup**：多副本部署时，`runCleanupTasks` / `runRequestDetailCleanup` 只在 leader 实例上跑（活实例 ID 排序最小者）。原来 6 副本同时清理会把共享 RDS 的 IOPS 放大 6 倍互相 race。`Coordinator=nil`（单实例）退化为"总是 leader"，行为不变。

2. **v14 migration：MySQL 索引列序**：v13 的 `(created_at, id)` 复合索引在 MySQL 上 EXPLAIN filtered 仅 ~10%（`WHERE status IN AND dev_mode = 0` 要回表）。改成 `(status, dev_mode, created_at, id)` 后 filtered ~99%，cleanup SELECT 从 30-70s 降到亚秒。沿用 v13 的 500k 行阈值 + manual SQL fallback。SQLite 用 partial index 已最优，不变。

3. **Dialect-aware batch params**：`ClearDetailOlderThan` 的 batchSize/sleep 按 dialect 区分：SQLite 保持 200/50ms（保护单写 WAL 锁），MySQL 改为 1000/20ms（行锁无单写阻塞，配合 v14 索引把吞吐拉到 ~5×）。

## Test plan
- [x] `go build ./...`
- [x] `go test ./...` 全绿（含 `tests/multiinstance`）
- [ ] 多副本部署上观察 leader 单跑日志、非 leader 跳过
- [ ] MySQL 8 上 EXPLAIN cleanup SELECT 走 `idx_proxy_requests_detail_cleanup_v2`
- [ ] 大表（>500k 行）升级路径打印 manual SQL 而非阻塞

## 发布说明

* **新增功能**
  * 添加了多实例部署下的清理任务协调机制，确保清理操作仅在主实例执行，避免重复处理。

* **优化改进**
  * 优化了数据库索引结构以提升清理任务性能。
  * 根据数据库类型动态调整批处理参数，改善大规模数据处理效率。

[Review Change Stack](https://app.coderabbit.ai/change-stack/awsl-project/maxx/pull/566?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/cmd/maxx/main.go
+++ b/cmd/maxx/main.go
@@ -329,6 +329,7 @@ func main() {
 		Settings:           settingRepo,
 		AntigravityTaskSvc: antigravityTaskSvc,
 		CodexTaskSvc:       codexTaskSvc,
+		Coordinator:        coord,
 	})
 
 	// Ensure default tenant exists
--- a/internal/core/task.go
+++ b/internal/core/task.go
@@ -3,10 +3,12 @@ package core
 import (
 	"context"
 	"log"
+	"sort"
 	"strconv"
 	"sync"
 	"time"
 
+	"github.com/awsl-project/maxx/internal/coordinator"
 	"github.com/awsl-project/maxx/internal/domain"
 	"github.com/awsl-project/maxx/internal/repository"
 	"github.com/awsl-project/maxx/internal/repository/sqlite"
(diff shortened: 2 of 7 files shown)


Bowl42:
Round 2 — review fixes pushed (e64ea1a):

**Addressed:**
- ✅ v14 `CREATE INDEX` now has `ALGORITHM=INPLACE, LOCK=NONE` (fail-fast on COPY fallback) — Claude
- ✅ v14 manual-SQL log softened: notes `DROP INDEX` is conditional on v13 having applied — Claude
- ✅ `checkDetailCleanupIndexHealth()` runs at startup on MySQL, warns loudly with manual SQL if no detail-cleanup index exists — addresses Codex's threshold-skip + no-index concern
- ✅ Startup log distinguishes leader-gated vs single-instance mode — Claude
- ✅ Tests added: `isCleanupLeader` (nil/self-smallest/self-not-smallest/error/empty), non-leader skip, `detailCleanupBatchParams`

**Out of scope (separate issue):**
- ⚠️ Codex: `coordinator.HeartbeatInterval` config is ignored — `StartHeartbeat` (heartbeat.go:16) hard-codes `ttl/3`. Originated in #561; tracking separately.

**Pushed back on:**
- Codex's "batch=1000 with no index = O(N²)": with cursor-based pagination, planner cost without index is full-scan-per-batch regardless of size. Larger batch reduces number of batches, so batch=1000 is no worse than batch=200 in that pathological case. The real fix is making "no index" visible, which the startup health check does.

Bowl42:
@coderabbitai review

awsldev:
Blocking concern before merge:

The MySQL index-health fallback currently treats either `idx_proxy_requests_detail_cleanup` (v13, `(created_at, id)`) or `idx_proxy_requests_detail_cleanup_v2` (v14, `(status, dev_mode, created_at, id)`) as "index present":

```sql
index_name IN ('idx_proxy_requests_detail_cleanup', 'idx_proxy_requests_detail_cleanup_v2')
```

That misses the important threshold-skip case for large MySQL tables:

1. v13 may have successfully created the old `(created_at, id)` index.
2. v14 may skip building `_v2` because `proxy_requests` is above `detailCleanupIndexRowThreshold`.
3. Startup health check sees the old v13 index and clears `detailCleanupIndexMissing`.
4. `detailCleanupBatchParams("mysql")` uses the fast path `1000 / 20ms`, even though the PR description says the v13 index filtered only ~10% and the v14 reorder is the actual performance fix.

So a large table with only the old v13 index can silently run the aggressive MySQL batch shape instead of the conservative fallback, while still needing manual v14 index creation. That undercuts the safety story of the threshold-skip path.

Please distinguish the optimized v14 index from the old v13 index in the health check, for example:

- treat `_v2` as fast-path ready;
- treat only v13/no v2 as degraded: warn with the v14 manual SQL and use conservative `200 / 50ms` until `_v2` exists;
- keep "no index at all" as the existing stronger warning path.

Local validation I ran while reviewing: `go test ./internal/core ./internal/repository/sqlite` passes.

Bowl42:
@awsl233777 nice catch — fixed in ae5e90d. Health check now has three states:

- **v14 `_v2` present** → fast path (1000/20ms)
- **v13 only, no `_v2`** → degraded: warns with manual v14 SQL, falls back to conservative 200/50ms batch (the case you described)
- **neither** → loudest warning + conservative batch

The threshold-skip scenario you described (v13 built normally on a smaller-then-grown table + v14 skipped because it crossed the threshold) is exactly the path that now correctly stays on the safe batch.

awsl233777 (approved):
Approved after blocker follow-up: v13-only MySQL detail-cleanup index is now treated as degraded, v14 index enables fast path, required checks are green, and targeted local review tests passed.

PR #574, opened 2026-05-22
feat(pricing): version model_prices on Update to preserve historical snapshots

Bowl42:
## Summary

把 `ModelPriceRepository.Update` 改成版本化语义,让历史 attempt 的 `ModelPriceID` 永远反查得到当时的价格快照——为后续按当前价重算成本、定期同步 LiteLLM 等场景铺路。

## Background

旧实现 `Update` 用 GORM `Save` 原地覆盖同一行价格记录。但 `Attempt.ModelPriceID` 是当时写入时定下的外键,如果 admin 后来改了价,反查时拿到的是被改后的新价格,**历史成本审计/重算会失真**。

`GetCurrentByModelID`、`ListCurrentPrices`、`ResetToDefaults` 等读路径**早就**已经版本化(取 MAX(created_at)、过滤 deleted_at=0),只是 `Update` 没跟上,导致 schema 上看起来支持多版本但实际只有一行。

## Changes

**`internal/repository/sqlite/model_price.go`**

- **`Update()` 改写**:事务里执行 (a) 按 ModelID 找当前行 (b) 与新价比较,字段全相等时 no-op 短路(避免 UI 误点产生空版本)(c) 否则软删旧行 + INSERT 新行;新 ID 通过 `price.ID` 回传给调用方
- **`GetByID()` 去掉 `deleted_at = 0` 过滤**:这是历史 attempt 按 ID 反查价格快照的路径,必须能查到已被软删(=历史版本)的行;列表/当前价查询全部走 `ListCurrentPrices` / `GetCurrentByModelID`,不受影响
- 新增私有 `pricesEqual` 工具函数,覆盖 6 个计费字段 + 4 个 premium num/denom + Has1MContext + Context1MThreshold

**`internal/repository/sqlite/model_price_test.go`** (新文件)

- `TestUpdate_InsertsNewRow_PreservesHistory`:Create → Update 改价 → 旧 ID 反查到旧价、`ListCurrentPrices` 只见新行
- `TestUpdate_NoOpWhenUnchanged`:同字段重复 Update 不产生新行,ID 保持不变;再改一个字段就产生新版本

## Migration / 升级影响

**Schema 不变,零 migration 步骤**。

- 现有 `model_prices` 行自动成为各自 ModelID 的"当前行"(deleted_at=0、最新 created_at)
- 下次 admin 改价才进入版本化分支
- Calculator 缓存在每次 admin 写入后 `LoadFromDatabase` 自动刷新,无需特殊操作
- 已写入的 `attempt.cost` 字段不变(数值在 attempt 写入时已固化);只有主动触发"按当前价重算"才会走 ModelPriceID 反查路径

## Test plan

- [x] `go test ./internal/repository/sqlite/ -run 'TestUpdate_'` 新增两个测试通过
- [x] `go test ./internal/repository/sqlite/ ./internal/pricing/ ./internal/service/` 全包测试无回归
- [x] `go build ./internal/... ./cmd/...` 干净
- [x] `go vet ./internal/repository/sqlite/ ./internal/pricing/` 干净
- [ ] 部署到一个干净实例,在 admin UI 改一次 sonnet 价格,验证 `model_prices` 表新增一行 + 旧行 `deleted_at` 被设置
- [ ] 同一价格连续保存两次,验证只有 1 行(no-op 短路生效)

## Summary by CodeRabbit

* **Chores**
  * 引入基于版本的价格更新流程：保留历史记录并在变更时创建新版本，输入与当前相同则不重复写入，查询仅返回最新有效价格。
  * 统一“当前版本”排序语义以提高在批量写入时的确定性，并补充相关注释以明确删除后应返回 404。
* **Tests**
  * 新增测试，覆盖版本化更新、历史保留与“无变更不产生新版本”的场景。

[Review Change Stack](https://app.coderabbit.ai/change-stack/awsl-project/maxx/pull/574?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/internal/repository/sqlite/model_price.go
+++ b/internal/repository/sqlite/model_price.go
@@ -61,15 +61,16 @@ func (r *ModelPriceRepository) BatchCreate(prices []*domain.ModelPrice) error {
 	return nil
 }
 
-// GetByID 获取指定ID的价格记录。
+// GetByID 获取指定ID的价格记录(仅未软删的)。
 //
-// 不过滤 deleted_at:每次 Update 都会软删旧行并插入新行(版本化),
-// 历史 attempt 的 ModelPriceID 指向已软删的旧行——重算/审计成本时仍需按
-// ID 精确取到当时的价格快照。需要"当前价格"请走 GetCurrentByModelID 或
-// ListCurrentPrices。
+// admin 单条编辑/查看路径,语义是"当前行";Delete 后再 GET 必须 404
+// (e2e: TestDeleteModelPrice)。"按历史 ModelPriceID 反查价格快照"目前
+// 没有实际调用方——RecalcFromAttempt 是用模型名走 GlobalCalculator 查
+// 当前价,不经此路径。如果未来需要历史反查,加一个独立方法,而不是
+// 削弱这里。
 func (r *ModelPriceRepository) GetByID(id uint64) (*domain.ModelPrice, error) {
 	var m ModelPrice
-	if err := r.db.gorm.First(&m, id).Error; err != nil {
+	if err := r.db.gorm.Where("deleted_at = 0").First(&m, id).Error; err != nil {
 		return nil, err
 	}
 	return r.toDomain(&m), nil
--- a/internal/repository/sqlite/model_price.go
+++ b/internal/repository/sqlite/model_price.go
@@ -1,8 +1,11 @@
 package sqlite
 
 import (
+	"errors"
 	"strings"
 	"time"
+
+	"gorm.io/gorm"
 
 	"github.com/awsl-project/maxx/internal/domain"
 	"github.com/awsl-project/maxx/internal/pricing"
@@ -58,10 +61,15 @@ func (r *ModelPriceRepository) BatchCreate(prices []*domain.ModelPrice) error {
 	return nil
 }
 
-// GetByID 获取指定ID的价格记录
+// GetByID 获取指定ID的价格记录。
+//
+// 不过滤 deleted_at:每次 Update 都会软删旧行并插入新行(版本化),
+// 历史 attempt 的 ModelPriceID 指向已软删的旧行——重算/审计成本时仍需按
+// ID 精确取到当时的价格快照。需要"当前价格"请走 GetCurrentByModelID 或
+// ListCurrentPrices。
 func (r *ModelPriceRepository) GetByID(id uint64) (*domain.ModelPrice, error) {
 	var m ModelPrice
-	if err := r.db.gorm.Where("deleted_at = 0").First(&m, id).Error; err != nil {
+	if err := r.db.gorm.First(&m, id).Error; err != nil {
 		return nil, err
 	}
 	return r.toDomain(&m), nil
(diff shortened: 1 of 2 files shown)


awsldev:
Review note from checking the current failed gate:

The versioning direction looks right, but the current implementation changes the semantics of the general `ModelPriceRepository.GetByID(id)` path by removing the `deleted_at = 0` filter. That makes historical snapshots readable, but it also leaks soft-deleted model prices back through normal API/admin reads.

Evidence from the current failing `e2e` check:

```text
model_prices_test.go:157: Expected status 404, got 200, body: {"id":1,..."modelId":"delete-test-model",...}
--- FAIL: TestDeleteModelPrice
```

So after deleting a model price, fetching it by ID now returns `200` instead of `404`.

Suggested fix: keep the public/current `GetByID` semantics filtered to non-deleted rows, and add a separate explicit method for historical lookups, e.g. `GetHistoricalByID` / `GetByIDIncludingDeleted`, used only by cost audit / attempt replay paths that need to resolve an old `ModelPriceID` snapshot.

That keeps the historical-cost contract without weakening the normal delete/read contract.

Bowl42:
Thanks for catching that — independently hit the same issue right after pushing and reverted in [`e3004c0`](https://github.com/awsl-project/maxx/pull/574/commits/e3004c0376ac726dc65a39cb2b8984e4d2502d18) just before your comment landed. The fix matches your suggestion exactly: `GetByID` is back to filtering `deleted_at = 0` (so admin Delete still 404s correctly), and the versioning contract is now verified via an `Unscoped()` raw-table read in the test instead.

On the "add a separate `GetHistoricalByID` for cost audit / attempt replay" point — I looked through current callers and there's actually no path today that needs it: `RecalcFromAttempt` resolves prices via model name through `GlobalCalculator()` (not via stale `ModelPriceID`), so the historical-by-ID lookup has zero callers right now. I left a comment in `GetByID` flagging the convention — if/when we add a historical reverse-lookup feature (e.g. "show this attempt's snapshot price in the UI"), it should land as a new explicit method rather than relaxing this one.

CI re-running on `e3004c0` now.

awsldev:
Follow-up after `e3004c0`: the original `GetByID`/delete regression is fixed and checks are now green, but there is still one versioning-contract edge that looks worth fixing before merge.

`Update()` currently resolves the current row with:

```go
Where("model_id = ? AND deleted_at = 0", price.ModelID).
    Order("created_at DESC").First(&current)
```

while `ListCurrentPrices()` uses a `MAX(id)`-based current-row selection, and `Create` / `BatchCreate` can preserve caller-supplied `CreatedAt`. That means different repository methods can disagree on which row is the current version when a model has multiple non-deleted rows or non-monotonic timestamps. For a versioned price table, that inconsistency can make no-op detection, soft-delete targeting, and listing semantics diverge.

Suggested fix: define one stable current-version ordering and use it everywhere, e.g. `created_at DESC, id DESC` (or `id DESC` if ID is the intended source of truth), then align `Update`, `ListCurrentPrices`, and any create/batch paths that populate `CreatedAt` with that rule.

Also worth adding a direct regression assertion in `TestUpdate_InsertsNewRow_PreservesHistory`: after the raw `Unscoped()` check proves the old snapshot is physically retained, assert `repo.GetByID(originalID)` returns not found. That locks the exact contract we just fixed: historical rows remain in storage, but normal admin/API reads must not expose them.

Bowl42:
Both points addressed in [`5a3fd39`](https://github.com/awsl-project/maxx/pull/574/commits/5a3fd39):

**1. Unified current-version ordering → `id DESC`**

Switched `Update`, `GetCurrentByModelID` (both exact + prefix branches), and `ListByModelID` from `Order("created_at DESC")` to `Order("id DESC")`, aligning with `ListCurrentPrices`' existing `MAX(id)`. Reasoning matches your call-out: `BatchCreate` (seed + `ResetToDefaults`) gives an entire batch the same `time.Now()` timestamp, so `created_at` is not a stable tiebreaker, while `id` is autoincrement + unique + strictly monotonic. Added a docstring on `GetCurrentByModelID` explaining the invariant so future contributors don't drift back.

**2. `GetByID(originalID)` not-found assertion**

Added to `TestUpdate_InsertsNewRow_PreservesHistory` right after the `Unscoped()` physical-retention check. The test now locks both halves: historical row physically retained in storage + invisible to normal admin reads, mirroring `TestDeleteModelPrice`'s 404 contract.

Local: `go test ./internal/repository/sqlite/ ./tests/e2e/` clean.

awsl233777 (approved):
Approved after re-checking the latest state:

- The original `GetByID` / soft-delete regression was fixed in `e3004c0`.
- The follow-up current-version consistency issue was addressed in `5a3fd39` by aligning current-version selection to `id DESC` and documenting the invariant.
- The regression assertion now verifies that the old soft-deleted snapshot is retained physically but not exposed through `GetByID`.
- Current checks are green: Backend Checks, Frontend Checks, e2e, playwright, and CodeRabbit has no actionable comments.

This looks ready from the review gate side.

PR #581, opened 2026-05-28
fix(usage): correctly extract OpenAI chat-completions + banana image tokens

Bowl42:
## Why

While verifying billing for `gemini-2.5-flash-image` ("nano-banana") through maxx I traced a request that returned HTTP 200 with a real image and upstream `usage.prompt_tokens=7, usage.completion_tokens=1290, usage.completion_tokens_details.image_tokens=1290`, but landed in the DB with `inputTokenCount=0 outputTokenCount=0 cost=0 modelPriceId=0`. The token extractor was silently zeroing it.

Two stacked bugs in `internal/usage/extractor.go`:

1. **Dispatch ordering** — `extractUsageFromMap` unconditionally ran `extractClaudeUsage` for any top-level `{"usage": …}`. Claude's extractor only reads `input_tokens` / `output_tokens`, so OpenAI chat-completions (`prompt_tokens` / `completion_tokens`) were dropped. The OpenAI branch further down was dead code, unreachable while `data["usage"]` is set. Affects all openai-clientType chat-completions traffic — Codex (Response API, `response.usage`) is unaffected because it goes through a different branch.

2. **Within `extractOpenAIUsage`** — naive last-wins overwrite of `prompt_tokens` by `input_tokens`. OpenAI-compatible aggregators (linkapi, oneapi, …) dual-emit Claude-style zero-padded `input_tokens:0` / `output_tokens:0` alongside the real OpenAI numbers for Claude-client compatibility. Even with #1 fixed, the zero pads would still clobber the real values.

Plus a third gap blocking image-rate billing: `applyImageTokenDetails` only knew the Images API naming (`input_tokens_details.image_tokens`, `output_tokens_details.image_tokens`). Banana goes through chat completions and uses `prompt_tokens_details.image_tokens` / `completion_tokens_details.image_tokens` instead.

## What

- `extractUsageFromMap` detects format from the usage map itself via a small `isOpenAIUsage` helper (presence of `prompt_tokens` or `completion_tokens`) and dispatches before the Claude fallback. Removes the now-redundant `choices`+`usage` block at the bottom (that path was always unreachable).
- `extractOpenAIUsage` switches `input_tokens` / `output_tokens` to an `else if` fallback so chat-completions naming wins when both are present.
- `applyImageTokenDetails` accepts both naming conventions.

## Tests

Three new tests in `extractor_test.go`:
- `TestExtractFromResponse_OpenAIChatCompletions` — plain chat completions
- `TestExtractFromResponse_OpenAIChatCompletionsZeroPadded` — linkapi-style dual emit (the actual production trigger)
- `TestExtractFromResponse_GeminiFlashImageViaChatCompletions` — banana with `completion_tokens_details.image_tokens=1290` correctly populates `OutputImageTokens`

All existing extractor tests (gpt-image-2 generations + edits, Claude / Codex / Gemini variants) untouched and still pass; full `go test ./internal/...` is green.

## Verification (live)

Traced via a maxx admin attempt for a banana request through linkapi:
- Before: `tokens=0/0 cost=0 modelPriceId=0`
- After (locally with this fix + a `gemini-2.5-flash-image` `model_prices` entry priced at Google list rates): tokens read `7/1290` with `OutputImageTokens=1290`, so the calculator bills the 1290 output tokens at the image rate ($30/M) rather than the text rate ($2.50/M) — ~12× difference per banana image.

## 发行说明

* **错误修复**
  * 修复了OpenAI兼容API转发时token计数被零覆盖导致计费异常的问题
  * 改进了图片token的提取逻辑

* **测试**
  * 新增单元测试用于验证OpenAI聊天完成API的token计数准确性

[Review Change Stack](https://app.coderabbit.ai/change-stack/awsl-project/maxx/pull/581?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/internal/usage/extractor.go
+++ b/internal/usage/extractor.go
@@ -129,8 +129,18 @@ func extractFromSSE(body string) *Metrics {
 // extractUsageFromMap extracts usage metrics from a parsed JSON map.
 // Handles multiple API formats.
 func extractUsageFromMap(data map[string]interface{}) *Metrics {
-	// Try Claude/Anthropic format: { "usage": { ... } }
+	// Try top-level { "usage": { ... } }. The same key serves both OpenAI chat
+	// completions (prompt_tokens / completion_tokens) and Claude / OpenAI Images
+	// (input_tokens / output_tokens), so we dispatch by inspecting the usage
+	// map itself. OpenAI-compatible relays (linkapi, oneapi, etc.) sometimes
+	// emit *both* schemas in the same payload — typically real numbers in
+	// prompt_tokens/completion_tokens and zero-padded input_tokens/output_tokens
+	// for Claude-client compatibility. Detecting OpenAI first prevents those
+	// zero-pads from silently zeroing chat-completions billing.
 	if usage, ok := data["usage"].(map[string]interface{}); ok {
+		if isOpenAIUsage(usage) {
+			return extractOpenAIUsage(usage)
+		}
 		return extractClaudeUsage(usage)
 	}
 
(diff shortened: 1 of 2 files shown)


awsldev:
Review note after checking the current PR state:

I reviewed the usage extraction changes and did not find a blocking issue in this patch.

What I checked:
- `internal/usage/extractor.go`
- `internal/usage/extractor_test.go`
- Current PR checks: Backend, Frontend, e2e, playwright, and CodeRabbit are all passing.
- CodeRabbit currently reports no actionable comments.
- `go test ./internal/usage` passes locally.

The implementation now prefers OpenAI-style `prompt_tokens` / `completion_tokens` when present, while keeping `input_tokens` / `output_tokens` as fallback. That matches the intended fix for mixed/zero-padded relay usage payloads. The added tests cover the OpenAI chat-completions path and image token detail extraction.

No code blocker from my side at this point. The remaining gate appears to be review approval.

awsl233777 (approved):
Approved after checking current diff, passing required checks, and no remaining actionable review blockers.

PR #582, opened 2026-05-28
fix(provider-proxy): record attempts + bill /provider/<id>/... direct dispatch

Bowl42:
## Why

`ProviderProxyHandler.directDispatch` — the path that serves `/provider/<id>/...` requests by talking to one provider directly, bypassing route selection — calls `adapter.Execute` bare, without any of the attempt-recording or cost-finalization the retry-driven dispatch middleware performs per iteration.

The visible damage in production: a provider-prefixed request landed in the DB with `proxyUpstreamAttemptCount=0`, **zero attempt rows**, `cost=0 multiplier=0 modelPriceId=0` — silently zero-billing every call on this URL prefix regardless of upstream usage. Sibling `/project/<slug>/...` and root-level traffic go through the dispatch middleware and are correctly billed; only the provider-prefix path was affected.

## What

Add `executor.ExecuteOnce`, a small exported helper that wraps a single `adapter.Execute` call with the per-attempt lifecycle: insert `ProxyUpstreamAttempt` row, wire `flow.KeyUpstreamAttempt` + `flow.KeyEventChan` so the adapter can emit usage events, spawn `processAdapterEventsRealtime` to drain them, run `pricing.FinalizeAttemptCost` on completion (success or failure), and on success mirror cost + `FinalProxyUpstreamAttemptID` + TTFT up to the `ProxyRequest` via `pricing.MirrorCostToRequest`.

`directDispatch` routes its `adapter.Execute` through `ExecuteOnce`, keeping its existing modifier / capture writer chain intact. If `executor` is nil (test-only wiring) it falls back to a bare adapter call rather than panicking — billing stays unrecorded in that path, matching prior behaviour.

## Tests

`internal/executor/single_attempt_test.go`:
- `TestExecuteOnce_PersistsAttemptAndMirrorsBilling` — fake adapter emits `EventMetrics(in=7, out=1290)`; asserts attempt row created with `Status=IN_PROGRESS` then updated to `COMPLETED`, multiplier `64000` (from `clientMultiplier.openai`), `proxyReq.ProxyUpstreamAttemptCount=1`, mirrored tokens / multiplier / `FinalProxyUpstreamAttemptID`.
- `TestExecuteOnce_RecordsFailedAttemptWithoutMirroringCost` — failed adapter still records `Status=FAILED` attempt, but `proxyReq.Cost` / `FinalProxyUpstreamAttemptID` stay zero (we don't bill failed upstream calls).

Full `go test ./internal/...` is green.

## Summary by CodeRabbit

* **新功能**
  * 为直连请求引入了单次尝试的记录与管理，覆盖尝试生命周期、状态与计费镜像
  * 在无执行器配置时改为返回服务器错误，避免绕过尝试记录的直接执行路径

* **修复/改进**
  * 统一了成功/失败路径的费用结算与镜像行为，确保计费一致性

* **测试**
  * 新增覆盖成功、创建失败、最终更新失败与适配器失败等场景的单元测试，验证持久化与计费逻辑

[Review Change Stack](https://app.coderabbit.ai/change-stack/awsl-project/maxx/pull/582?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/internal/executor/single_attempt.go
+++ b/internal/executor/single_attempt.go
@@ -68,7 +68,14 @@ func (e *Executor) ExecuteOnce(
 		attempt.TenantID = maxxctx.GetTenantID(c.Request.Context())
 	}
 	if err := e.attemptRepo.Create(attempt); err != nil {
+		// Fail fast: if the attempt row can't be persisted, downstream cost
+		// mirroring would write a billing row whose backing attempt doesn't
+		// exist — exactly the desync this PR is meant to prevent. The caller
+		// surfaces this as a normal failure path on the ProxyRequest, leaving
+		// proxyReq.ProxyUpstreamAttemptCount at 0 (no phantom-incremented
+		// counter).
 		log.Printf("[Executor] ExecuteOnce: failed to create attempt record: %v", err)
+		return nil, err
 	}
 
 	proxyReq.ProxyUpstreamAttemptCount++
--- a/internal/executor/single_attempt.go
+++ b/internal/executor/single_attempt.go
@@ -118,7 +118,15 @@ func (e *Executor) ExecuteOnce(
 		attempt.RequestInfo = nil
 		attempt.ResponseInfo = nil
 	}
-	_ = e.attemptRepo.Update(attempt)
+	// Persisting the final attempt is best-effort here: adapter.Execute may
+	// already have written the upstream response to the client, so converting
+	// a late DB update failure into a returned adapter error could append an
+	// error body onto an otherwise successful response. Keep the request billing
+	// mirror below as the durable billing record; stale IN_PROGRESS attempts are
+	// handled by the existing stale-attempt cleanup path.
+	if updateErr := e.attemptRepo.Update(attempt); updateErr != nil {
+		log.Printf("[Executor] ExecuteOnce: failed to update final attempt record: %v", updateErr)
+	}
 	if e.broadcaster != nil {
 		e.broadcaster.BroadcastProxyUpstreamAttempt(attempt)
 	}
--- a/internal/executor/single_attempt.go
+++ b/internal/executor/single_attempt.go
@@ -1,0 +1,126 @@
+package executor
+
+import (
+	"log"
+	"time"
+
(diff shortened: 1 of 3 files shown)


Bowl42:
@coderabbitai review

awsldev:
Review note from checking the current patch:

The direction is right: direct dispatch now goes through `Executor.ExecuteOnce`, attempt creation is fail-fast, and the missing-executor fallback no longer silently bypasses billing. I also verified the focused tests pass:

```text
go test ./internal/executor ./internal/handler ./internal/pricing
```

One remaining concern before approval: `internal/executor/single_attempt.go` still ignores the final attempt persistence error:

```go
_ = e.attemptRepo.Update(attempt)
```

That happens before the success path mirrors `FinalProxyUpstreamAttemptID` and billing fields back to `proxyReq`. If the final attempt update fails after adapter execution, the parent request can still be persisted as completed/billed while the backing attempt row remains stale (`IN_PROGRESS`, missing final status/cost/details). That is very close to the billing/audit drift this PR is trying to eliminate for direct dispatch.

Can we either handle that update error before mirroring to `proxyReq`, or explicitly document why best-effort final attempt persistence is acceptable for this direct-dispatch billing path?

awsldev:
Follow-up after the review concern:

Pushed `12f5bc4` to document and pin the final `attemptRepo.Update(attempt)` behavior.

What changed:
- Added an explicit code comment explaining why the final attempt update remains best-effort in direct dispatch: `adapter.Execute` may already have written the upstream response, so returning a late DB update error could append an error body to an otherwise successful client response.
- Added logging for a failed final attempt update instead of silently discarding it.
- Added `TestExecuteOnce_FinalUpdateFailureStillMirrorsBilling` to document the contract that the parent request billing mirror remains the durable billing record while stale attempt cleanup handles stale `IN_PROGRESS` attempts.

Validation:

```text
go test ./internal/executor ./internal/handler ./internal/pricing
```

Current PR checks are all passing.

awsl233777 (approved):
Approved after resolving the direct-dispatch billing review concern, documenting the final attempt update behavior, adding regression coverage, and passing required checks.

PR #584, opened 2026-05-28
feat(routing): weighted load balancing with session-affinity sticky

Bowl42:
## Summary

Adds real weighted load balancing for routes plus a Redis-backed session-affinity layer on top. Previously routing was either strict priority (sequential fallback) or a per-request reshuffle whose `Route.Weight` was never exposed in the UI.

- **Weighted distribution**: `Route.Weight` is editable from the routes form and used by the existing `weighted_random` strategy. The shuffle is now seeded from `(APITokenID, SessionID)` so the same session sees a stable fallback order while different sessions diverge — implicit affinity, even before sticky is enabled.
- **Session-affinity sticky** (`internal/sticky/`): a new package storing `(session → provider)` bindings in Redis (memory fallback for standalone) with a `policyFingerprint` embedded in the key so any route config change auto-invalidates every binding. TTL configurable (default 30 min). Scope toggles between `token` (one binding per API token, coarser, best prompt-cache locality) and `conversation` (per session id).
- **Architecture** mirrors `internal/cooldown`: `Store` interface, Redis + memory impls, `sticky.Default()` singleton wired in `core/coordinator_setup.go`, zero new shared-state dependencies.
- **Router.Match** now returns `MatchResult{Routes, Sticky}`; the dispatcher SETEXes the binding only after a confirmed-good upstream response. Failure paths intentionally don't branch by error class — failover + overwrite-on-success handles 4xx/429/5xx uniformly; the next request's `Match` re-filters via cooldown, which is already distributed.
- **Pre-existing fix bundled**: `AdminService.GetRoutingStrategy` was calling `repo.GetByProjectID` with the strategy ID, causing PUT `/admin/routing-strategies/{id}` to 404 for any global strategy. Adds a real `GetByID` to the repository interface + sqlite + cached layer.

UI exposes the sticky config block in the routing-strategies page only when `weighted_random` is selected (priority writes `config: null`).

## Test plan

End-to-end verification was run against the full HTTP proxy stack with three mock upstreams; all probes are committed as tests.

- [x] **Phase A — large-sample weight distribution.** 2000 distinct sessions with weights 7/2/1 → 70.2% / 19.3% / 10.4% (each bucket within ±3% of expected). `TestVerifyRoutingPhase2`
- [x] **Phase B — seeded determinism.** Same session repeated 10× with sticky OFF lands on a single provider every time. `TestVerifyRoutingPhase2`
- [x] **Phase C — sticky write+read.** Warm-up + 30 follow-ups for one session all hit the primed provider; SETEX-on-success + GET-on-Match exercised through the real dispatcher. `TestVerifyRoutingPhase2`
- [x] **Phase D — cooldown failover.** Cooldown the primed provider; 20 same-session requests transparently land elsewhere with the pinned provider receiving 0 hits. `TestVerifyRoutingPhase2`
- [x] **Failover spread probe.** 200 distinct sessions under cooldown spread across surviving providers per their 2:1 weight ratio. `TestVerifyRoutingPhase2`
- [x] **Error-class behavior.** Flipping a mock to return 500: dispatcher fails over to the healthy provider, sticky overwrites to it, and after the original recovers there is no drift back. Client sees 200 throughout. `TestVerifyRoutingErrorClass`
- [x] **Multi-instance Redis consistency.** Two `redisStore` instances backed by the same miniredis: cross-instance read after write, overwrite, TTL expiry (FastForward), key escaping with colons/spaces, 50 concurrent writers under `-race`, cross-instance delete, raw uint64 schema. `TestRedisStoreCrossInstance`
- [x] **Full regression sweep.** `go test ./internal/... ./tests/...` green.

## Notes

- Single-session fallback is deterministic (Phase D one session always picked p3, not 2:1) — that's the seeded-shuffle property by design; aggregate failover across many sessions still spreads by weight (see probe). Documented in the test log.
- Codex was consulted twice during design; final architecture (weighted_random + lazy sticky with policy fingerprint, instead of consistent hashing) and the production-PASS bar (large sample, error class, 2-node Redis) both came out of that review.

## Summary by CodeRabbit

* **新功能**
  * 为加权随机路由添加会话粘性（sticky），支持按 API Token 或会话绑定提供者并可配置 TTL；路由支持权重配置。
* **修复/改进**
  * 路由匹配在同一会话中更确定稳定，粘性信息可跨实例持久化并在请求成功后回写，写失败不影响主请求流；管理端可按策略 ID 查询。
* **测试**
  * 新增端到端与存储一致性测试覆盖粘性、故障切换与冷却场景。
* **文档**
  * 增加 UI 文案与环境变量说明（可选共享种子）。

[Review Change Stack](https://app.coderabbit.ai/change-stack/awsl-project/maxx/pull/584?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -6,3 +6,11 @@ AWS_REGION=us-east-1
 # Bedrock-compatible proxy (e.g. code0, bedrock-access-gateway)
 BEDROCK_PROXY_BASE_URL=
 BEDROCK_PROXY_API_KEY=
+
+# Routing — optional shared salt for the weighted_random strategy.
+# Set the same value on every instance for consistent first-pick
+# behavior in multi-instance deployments. When unset, each process
+# generates a per-process random salt; anti-grinding and Redis sticky
+# still work, but the pre-sticky first pick for a given (token, session)
+# may differ across instances until the sticky binding is written.
+MAXX_ROUTING_SEED_SALT=
--- a/README.md
+++ b/README.md
@@ -268,6 +268,7 @@ codex
 | `MAXX_ADMIN_PASSWORD` | Enable admin authentication with JWT. Default username: `admin`, password: the value of this variable |
 | `MAXX_DSN` | Database connection string |
 | `MAXX_DATA_DIR` | Custom data directory path |
+| `MAXX_ROUTING_SEED_SALT` | Optional shared secret for the `weighted_random` routing strategy. If unset, each process generates its own random salt — anti-grinding still holds and Redis sticky bindings still converge after the first successful request, but the pre-sticky first-pick order for the same `(token, session)` can differ across instances. Set the **same value on every instance** when you need consistent first-pick behavior in multi-instance deployments. |
 
 ### System Settings
 
--- a/README_CN.md
+++ b/README_CN.md
@@ -268,6 +268,7 @@ codex
 | `MAXX_ADMIN_PASSWORD` | 启用管理员 JWT 认证。默认用户名：`admin`，密码为该变量的值 |
 | `MAXX_DSN` | 数据库连接字符串 |
 | `MAXX_DATA_DIR` | 自定义数据目录路径 |
+| `MAXX_ROUTING_SEED_SALT` | 可选共享密钥，用于 `weighted_random` 路由策略。未设置时每个进程会生成自己的随机盐——防 SessionID 枚举仍然成立，Redis sticky 绑定也会在首次成功后跨实例收敛；但相同 `(token, session)` 在 sticky 写入前的首选顺序在各实例间可能不一致。**多实例部署且需要一致首选顺序**时，请在所有实例上设置相同的值 |
 
 ### 系统设置
 
--- a/docs/multi-instance-rfc.md
+++ b/docs/multi-instance-rfc.md
@@ -73,10 +73,31 @@ boot, you keep getting Redis errors until Redis recovers."
 | `MAXX_PROXY_REQUEST_SWEEP_INTERVAL`         | `45s`         | stale-request sweep period                        |
 | `MAXX_COOLDOWN_GEN_SYNC_INTERVAL`           | `2s`          | per-provider generation re-check throttle         |
 | `MAXX_COORDINATOR_RECONNECT_INTERVAL`       | `5s`          | degraded-mode reconnect cadence                   |
+| `MAXX_ROUTING_SEED_SALT`                    | `""`          | shared HMAC salt for weighted_random first-pick   |
 
 Legacy `MAXX_REDIS_URL` is retained as an alias for
 `MAXX_COORDINATOR_REDIS_URL` for backward compatibility with the original
 draft of this work.
+
+`MAXX_ROUTING_SEED_SALT` is **optional but recommended** in multi-instance
+deployments using the `weighted_random` routing strategy. Each instance
+falls back to a per-process random 32-byte salt when the variable is
+unset. Anti-grinding (a client cannot brute-force `X-Session-Id` to
+steer themselves onto a specific upstream) still holds in either mode,
+because the salt is never exposed. Redis sticky bindings remain
+consistent across instances as soon as the first dispatch succeeds —
+sticky reads the provider ID directly from Redis without re-shuffling.
+The visible difference between "shared salt" and "per-process salt" is
+only the **pre-sticky first pick** for the same `(token, session)`:
+without a shared salt, instance A and instance B may pick a different
+provider on the first request; whichever instance's pick succeeds first
+writes the sticky binding, and both instances converge for the rest of
+the session's lifetime.
+
+Set the same `MAXX_ROUTING_SEED_SALT` value on every instance if you
+want first-pick consistency before sticky writes land — useful for
+debugging hot-spot reports or making cold-start behavior reproducible
+across the fleet.
 
 ## Stale request reclamation
 
--- a/go.mod
+++ b/go.mod
@@ -3,6 +3,7 @@ module github.com/awsl-project/maxx
 go 1.26.0
 
 require (
+	github.com/alicebob/miniredis/v2 v2.38.0
 	github.com/andybalholm/brotli v1.2.0
 	github.com/aws/aws-sdk-go-v2 v1.41.5
 	github.com/aws/aws-sdk-go-v2/credentials v1.19.14
@@ -30,7 +31,6 @@ require (
 require (
 	cloud.google.com/go/compute/metadata v0.3.0 // indirect
 	filippo.io/edwards25519 v1.1.1 // indirect
-	github.com/alicebob/miniredis/v2 v2.38.0 // indirect
 	github.com/aws/smithy-go v1.24.2 // indirect
 	github.com/bep/debounce v1.2.1 // indirect
 	github.com/bytedance/gopkg v0.1.3 // indirect
--- a/internal/core/coordinator_setup.go
+++ b/internal/core/coordinator_setup.go
@@ -8,6 +8,7 @@ import (
 	"github.com/awsl-project/maxx/internal/cooldown"
 	"github.com/awsl-project/maxx/internal/coordinator"
 	"github.com/awsl-project/maxx/internal/repository/cached"
+	"github.com/awsl-project/maxx/internal/sticky"
 )
 
 // CoordinatorComponents 是 SetupCoordinator 的输出。
@@ -50,6 +51,7 @@ func SetupCoordinator(parentCtx context.Context, instanceID string, forceStandal
 	coordinator.StartHeartbeat(coordCtx, coord, cfg.InstanceTTL)
 
 	cooldown.Default().SetCoordinator(coordCtx, coord)
+	sticky.Default().SetCoordinator(coordCtx, coord)
 
 	cleanup := func() {
 		// 顺序很重要:先 Unregister 让其他实例感知到本实例下线,
--- a/internal/domain/model.go
+++ b/internal/domain/model.go
@@ -625,7 +625,23 @@ var (
 type RoutingStrategyConfig struct {
 	// 加权随机策略的权重配置等
 	// 根据具体策略扩展
+
+	// Sticky / session-affinity 配置（用于 weighted_random 策略；priority 下忽略）
+	// 启用后：同一 (api token[+session]) 命中过的 provider 会在 TTL 内被记住，
+	// 后续请求优先尝试它（最大化上游 prompt cache 命中率）。
+	StickyEnabled    bool                 `json:"stickyEnabled,omitempty"`
+	StickyScope      RoutingStickyScope   `json:"stickyScope,omitempty"`      // "token" | "conversation"，默认 "token"
+	StickyTTLSeconds int                  `json:"stickyTTLSeconds,omitempty"` // 默认 1800（30 分钟），<=0 取默认
 }
+
+type RoutingStickyScope string
+
+const (
+	// 按 API token 粘性：同 token 的所有 session 都打同一 provider（命中率高，亲和粒度粗）
+	RoutingStickyScopeToken RoutingStickyScope = "token"
+	// 按 conversation 粘性：(token, sessionID) 粘性（亲和粒度细，sticky 项更多）
+	RoutingStickyScopeConversation RoutingStickyScope = "conversation"
+)
 
 // 路由策略
 type RoutingStrategy struct {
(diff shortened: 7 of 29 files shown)


awsldev:
Review note from checking the sticky routing path:

The sticky read path should probably be bound to the request context (or at least a short bounded timeout) instead of `context.Background()`.

Evidence in the current diff:

```go
// internal/router/router.go
if pinned, ok := sticky.Default().Get(context.Background(), key); ok {
    matched = promoteByProvider(matched, pinned)
}
```

The write-back path already uses the request context from dispatch:

```go
sticky.Default().Set(ctx, state.stickyWrite.Key, matchedRoute.Provider.ID, state.stickyWrite.TTL)
```

Likely issue: if Redis/coordinator is slow or wedged, the route-matching request cannot be cancelled by the client/request deadline while waiting on the sticky GET. This also happens while `Router.Match` still holds the router read lock, so a stuck Redis read can unnecessarily delay router mutations/reconfiguration.

Suggested fix: carry `context.Context` into `Router.Match` / `MatchContext` from the proxy request and use that for `sticky.Default().Get`, preferably with a small timeout if the request context has no deadline. That keeps sticky best-effort and non-fatal without letting affinity lookup become an unbounded request-path dependency.

Bowl42:
## Round 1 review responses — `a408fa0`

Aggregated must-fix from three independent reviewers (local Claude subagent, codex CLI, CodeRabbit) and applied:

### 🔴 Must-fix (applied)

| # | Source | Finding | Resolution |
|---|---|---|---|
| 1 | Codex+Claude | `policyFingerprint` 48-bit → birthday-bound adversarial | Bumped to **128 bits** (`sum[:16]`) |
| 2 | Codex+Claude | `(APITokenID, SessionID)` seed predictable → traffic steering by SessionID grinding | HMAC with `MAXX_ROUTING_SEED_SALT` (env var; hardcoded fallback for personal deploys; required-shared for multi-instance) |
| 3 | Claude | `makeSessionSeed` time fallback when `apiTokenID=0` broke Phase B determinism for unauth setups | Seed now anchors to **routing context** (tenant/client/project) — never time |
| 4 | Claude | Sticky `Set(state.ctx, ...)` — streaming ctx is `Done` by write path | Switched to fresh `context.WithTimeout(Background, 2s)`; read path was already Background, restoring symmetry |
| 5 | CodeRabbit | Sticky `Get(context.Background())` in `Match` had no timeout → Redis hiccup stalls match | Bounded with **100ms** timeout |
| 6 | CodeRabbit | `memoryStore.Get` expired entries never deleted → in-process leak under churn | Lazy delete under write-lock re-check |
| 7 | CodeRabbit | `store_redis_test.go` discarded `err` from `Get` → real failures masked as miss | All `Get` calls now assert `err == nil` first |
| 8 | CodeRabbit | `srvB.CloseClientConnections()` in ErrorClass test unreliable (listener stays open) | Refactored to symmetric primed/spare model with `setStatus(500)` injection on either side |

### 🟡 Should-fix (applied)

- `StickyTTLSeconds int → int64` (Claude — 32-bit platform overflow guard)
- `sticky.Manager.SetCoordinator` lifecycle expectation documented (Codex+Claude)
- Phase D probe assertion strengthened: previously only checked cooled-down provider got 0 hits; now bounds surviving providers' counts to ±20 of expected (133/67 for 200 sessions @ 2:1) (Claude)
- Explicit **429 sub-probe** added to `TestVerifyRoutingErrorClass`: pinned side flipped to 429 mid-test, asserts failover target + sticky overwrite (Codex)

### ❓ False positives (examined, not changed)

- Claude + CodeRabbit both flagged Phase D cooldown cleanup as incomplete. `cooldown.Manager.ClearCooldown(providerID, "", "")` treats empty clientType+model as "clear ALL variants for this provider" (`manager.go:402-413` — iterates and deletes every cooldown key matching providerID, plus calls `DeleteByProvider` on the Redis store). The single cleanup line already does what the suggested 3-line cleanup would have done. Not changed; added the manager-side guarantee to the test comment.
- Claude flagged `Manager.SetStore` as racy with in-flight `Get`/`Set`. The `atomic.Pointer[Store]` pattern is intentional: in-flight callers complete against the old backend, new callers hit the new one. **Documented** the constraint instead of changing the design — see the new comment on `SetCoordinator`.

### 🟢 Deferred (not blockers)

- Codex NIT: refactor tests to table-driven. Would lose per-phase log readability.
- Codex NIT: hot-key documentation. Marginal under LLM latencies (sticky GET ≪ upstream call); add later if observed.
- Claude 🟡: `GetByID` cache scan is O(N). Call rate is admin-UI-only.
- Claude 🟡: frontend grid responsive — purely cosmetic.

All probes + full e2e + multi-instance suites green locally after the fixes.

Re-requesting Round 2 review.

Bowl42:
@coderabbitai full review

Bowl42:
## Round 2 review responses — `a3bb839`

Round 2 verdicts: **Codex APPROVE_WITH_NITS** (1 must-fix to APPROVE), **Claude APPROVE_WITH_NITS** (3 nice-to-haves), **CodeRabbit 2 Major** outside-diff findings. All addressed:

### 🔴 Must-fix (applied)

| # | Source | Finding | Fix |
|---|---|---|---|
| 1 | Codex | Hardcoded salt fallback (`maxx/v1/routing-seed-salt/default`) = public predictable secret; anyone with the binary can grind sessions | Lazy resolution via `sync.Once`; fallback is now **32 bytes from `crypto/rand`**. Multi-instance still wants `MAXX_ROUTING_SEED_SALT` for cross-instance first-pick consistency, but the silent-default-public-salt path is gone |
| 2 | CodeRabbit (Major) | `r.mu.RLock` held across sticky `Get` (up to 100ms) blocks adapter refresh / removal on write side | Lock scoped to matched-set build only; released before sticky call |
| 3 | CodeRabbit (Major) | Phase D probe hardcodes `snapP[0]==0` and `~133/67` — with per-process random salt, primed provider is now random; assertion would fire on any `primedProvider != 0` | Probe computes expected ratio dynamically from `primedProvider` + `weights[]`, validates each surviving bucket against its own ±20 band |

### 🟡 Should-fix (applied)

| # | Source | Finding | Fix |
|---|---|---|---|
| 4 | Codex | Sticky Set timeout 2s too conservative for tail-latency | 2s → **500ms** |
| 5 | Codex | HMAC layout — strings without length prefix risk encoding ambiguity | Refactored to **uniform length-prefixed** uint64 LE encoding; no positional separators |
| 6 | Claude | `memoryStore.Get` TOCTOU: recheck preserves fresh entry but returns phantom miss | Returns the fresh `providerID` if a Set raced in during the lock upgrade |
| 7 | Claude | Sticky `Manager.Get` errors silently swallowed; chronic Redis outage produces zero signal | Rate-limited warning (≤1/min) via `atomic.Int64` timestamp |
| 8 | Claude | `routingSeedSalt` bound at package init — `t.Setenv` ineffective | Resolved by the `sync.Once` lazy pattern from #1 |

### 🟢 Nits applied

- `memoryStore.Get` recheck condition asymmetry (`After` vs `!Before`) → both use `time.Now().Before(expiresAt)` consistently after refactor

### ❓ Deferred / not applicable

- **CodeRabbit nit (round 1, re-raised round 2)**: Phase D cooldown cleanup. Confirmed again with `cooldown.Manager` source — `ClearCooldown(providerID, "", "")` enters the wildcard branch at `manager.go:402-413` and removes every cooldown variant + the Redis-store `DeleteByProvider` call. Per-variant clears would be redundant.
- Codex NIT (hot-key documentation): sticky GET is now bounded at 100ms and any chronic contention surfaces via the new rate-limited error log. Doc note no longer load-bearing.
- Claude nit (429 sub-probe assertion could tighten further): current `primed < reps-1` already catches the regression Codex was guarding against; tighter bound risks flakes on legitimate sticky transition.

Verified probe runs (multiple seeds via the new `crypto/rand` salt):
```
verify> Phase A — 2000 sessions: p1=1398 (69.9%) p2=400 (20.0%) p3=202 (10.1%)  ← near-perfect
verify> Phase D — cooldown on p1 (weight 7), 200 distinct sessions: p1=0 p2=129 p3=71
verify> ErrorClass 429 probe — pB→429, 10 requests: A=10 B=1 (expect majority pA)
```

Re-requesting round 3 review.

Bowl42:
@coderabbitai full review

Bowl42:
## Round 3 review responses — `ab58f14`

Round 3 verdicts:
- **Claude subagent:** APPROVE ✅ (no must-fix, two trivial nits applied below)
- **Codex CLI:** APPROVE_WITH_NITS (1 NIT applied; #2 deferred as opt-in config)
- **CodeRabbit:** awaiting r3 (will re-evaluate this push)

### Applied

| # | Source | Finding | Fix |
|---|---|---|---|
| 1 | Codex r3 | Phase D probe ±20 hardcoded → drifts to ~4σ at 7:2 ratio; future weight changes need retune | k·σ dynamic with k=4 per-bucket: `4·sqrt(N·p·(1-p))` |
| 2 | Claude r3 | crypto/rand fallback fills only 8 of 32 bytes → HMAC entropy collapses to 64 bits on that path | splitmix-stepped expansion across full 32-byte buffer |
| 3 | Claude r3 | Stale comment references `routingSeedSaltForTest` helper that doesn't exist | Removed |

### Deferred (acknowledged, not blocking)

- **Codex r3 NIT #2**: Make sticky Set 500ms timeout configurable for cross-region deployments. Reasonable in the future but the current 500ms is fine for typical deployments and the write is best-effort anyway; no point inflating config surface without an observed need.

### Verified runtime

```
verify> Phase A — 2000 sessions: p1=1392 (69.6%) p2=414 (20.7%) p3=194 (9.7%)
verify> Phase D — cooldown on p1 (weight 7), 200 distinct sessions: p1=0 p2=143 p3=57
        (expects p2≈133±4σ, p3≈67±4σ — both within band)
verify> ErrorClass 429 probe — pB→429, 10 requests: A=10 B=1 ✓
```

Two of three reviewers at APPROVE / APPROVE_WITH_NITS-already-applied. Re-requesting CodeRabbit for the final pass.

Bowl42:
@coderabbitai review

awsldev:
Deployment note from the follow-up review after `ab58f14`:

The routing seed salt fix looks directionally right, but `MAXX_ROUTING_SEED_SALT` currently appears only in code comments / startup logs. I could not find it in `.env.example`, `README*`, or `docs/multi-instance-rfc.md`.

For multi-instance deployments, this is worth documenting before/with merge:

- if unset, each process generates its own random salt;
- that preserves anti-grinding behavior and per-process determinism;
- Redis sticky bindings still converge after the first successful request;
- but the pre-sticky first-pick weighted order for the same `(token, session)` can differ across instances;
- operators who need consistent first-pick behavior across instances should set the same `MAXX_ROUTING_SEED_SALT` on every instance.

Suggested minimal fix: add the env var to `.env.example` and the multi-instance deployment docs as an optional-but-recommended shared secret for weighted_random + sticky deployments.

Bowl42:
@awsl233777 thanks for the two follow-up review notes — both addressed in `5e8298f`.

### 1. Request context plumbed into sticky Get ✅

`MatchContext` now carries the originating request's `context.Context`. `Router.Match`'s sticky lookup derives its 100ms timeout from that ctx instead of `context.Background()`, so a client cancel or a request-level deadline shortens the wait the same way it would for any other downstream IO. Nil Ctx is still accepted as a fallback so non-proxy call sites don't have to plumb anything.

```go
// internal/router/router.go
parentCtx := ctx.Ctx
if parentCtx == nil {
    parentCtx = context.Background()
}
stickyCtx, stickyCancel := context.WithTimeout(parentCtx, 100*time.Millisecond)
```

Also noting: the read lock on `r.mu` was already released before this point in round 2 (CodeRabbit Major #1), so a stuck sticky read can no longer delay adapter refresh / removal even with the new ctx wiring.

### 2. `MAXX_ROUTING_SEED_SALT` documented ✅

Added to:
- `README.md` and `README_CN.md` — environment-variables table
- `docs/multi-instance-rfc.md` — config table + a paragraph that explicitly calls out:
  - per-process random fallback when unset
  - anti-grinding still holds either way (salt never exposed)
  - Redis sticky bindings still converge after the first successful request
  - the only visible difference is the **pre-sticky first pick** for a `(token, session)` may differ across instances until the sticky write lands
- `.env.example` — entry with inline note pointing at the multi-instance implication

Full regression green; let me know if you want any wording tweaks on the docs.

Bowl42:
@coderabbitai fixed in `d3eb13a` — `Manager.Get` now filters `context.Canceled` and `context.DeadlineExceeded` via `errors.Is` before invoking the rate-limited error logger. Genuine Redis / store failures still surface; benign cancellations from the request-ctx wiring (or our own 100ms timeout cap) no longer drown the signal.

Tests green, including the e2e probes that now exercise the new ctx-derived sticky lookup.

awsl233777 (approved):
Approved after verifying the PR is open, mergeable, non-draft, and all required checks are passing.

PR #588, opened 2026-06-01
Fix Bedrock adaptive thinking retry

Bowl42:
## Summary
- detect Bedrock 400s that reject classic thinking.type=enabled
- rewrite classic thinking requests to adaptive shape with output_config.effort
- add focused tests for the adaptive rewrite and error matcher

## Tests
- go test ./internal/adapter/provider/...

## Summary by CodeRabbit

* **新增功能**
  * 请求重试机制扩展到最多四次，提升请求成功率。
* **修复与改进**
  * 遇到上游拒绝经典思维形式时，自动将思维参数改写为自适应形态并单次重放以恢复请求，避免重复触发重试。
  * 增强对相关错误的识别与处理，改善鲁棒性与故障恢复。
* **测试**
  * 新增改写与错误识别相关单元测试，覆盖多种场景。

### DIFF
--- a/internal/adapter/provider/bedrock/adapter.go
+++ b/internal/adapter/provider/bedrock/adapter.go
@@ -240,13 +240,16 @@ func (a *BedrockAdapter) Execute(c *flow.Ctx, provider *domain.Provider) error {
 	// Build upstream URL
 	upstreamURL := buildBedrockURL(region, bedrockModelID, clientWantsStream)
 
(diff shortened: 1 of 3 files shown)


awsl233777 (approved):
Approved after gate check: all required checks are green, CodeRabbit has no current actionable comments, and the prior thinking-object concern is covered by the current diff and regression tests.

PR #605, opened 2026-06-05
feat(routes): 在路由页展示生效策略并支持行内编辑权重

Bowl42:
## 背景

routes 页面(全局页和 project 页,二者共用 `ClientTypeRoutesContent`)一直存在两个缺口:

1. **看不到生效的路由策略** —— 站在某个 client/project 的路由列表前,无从得知走的是 `priority` 还是 `weighted_random`,策略配置和路由管理两个页面割裂。
2. **weight 不可见也不可编辑** —— 新建 route 时 `weight` 被硬编码为 1,列表里既不显示也无法修改,排序只能拖拽改 position。结果:即使把某 project 切到 `weighted_random`,纯前端用户也看不到它生效、更调不了权重(只能靠 CLI `route set-weight` 或 API)。

## 改动

- **策略横幅**:在路由列表顶部展示当前作用域**实际生效**的策略,解析顺序与后端 `router.getRoutingStrategy` 一致(project 专属 → 全局 → priority 默认),并标注「继承自全局」/「默认」,附带跳转到「路由策略」页的快捷入口。
- **行内权重编辑**:当生效策略为 `weighted_random` 时,每条路由行显示一个可编辑的权重输入框。失焦 / 回车提交,`<1` 归一到 1,值未变不发请求,Esc 取消;pointer 事件做了 `stopPropagation`,不会误触发拖拽或详情弹窗。复用已有的 `updateRoute` mutation(后端早已支持 weight 字段)。
- **会话亲和(sticky)状态**:`weighted_random` 下横幅额外显示 session 亲和是否开启及其 scope(token/conversation)和 TTL。亲和需要在策略 config 里**单独开启**(只切 `weighted_random` 不会自动生效 —— 见 `router.go` 的三段式 gating),过去 routes 页完全看不到它,导致"权重明明配了却看不出被亲和黏住"。现在一眼可见;priority 下不显示(那里亲和是 no-op)。
- 全局页与所有 project 页**自动同时生效**(共用同一组件,按 `projectID` 区分作用域)。
- `showWeight` 透传到 memo 化的行组件,并同步更新其 equality 比较函数,保持原有重渲染优化。
- 新增 en / zh 文案。

## 改动文件

- `web/src/components/routes/ClientTypeRoutesContent.tsx` —— 策略解析 + 横幅 + 传 `showWeight`
- `web/src/pages/client-routes/components/provider-row.tsx` —— `RouteWeightControl` 行内权重编辑 + prop 透传
- `web/src/locales/{en,zh}.json` —— 文案

## 验证

- `tsc` 类型检查通过
- `eslint` 干净
- `prettier` 通过(pre-commit lint-staged 已跑)
- `vite build` 生产构建成功

## 截图

> 截图取自本地实例 + 假数据(`Demo Provider Alpha/Bravo/...` + `api.example.com`),不含任何真实数据。

**`priority`(默认)** —— 横幅显示 `Priority (by position)` + `(default)`,提示拖拽排序,不显示权重:

routes priority

**`weighted_random`** —— 横幅切到 `Weighted Random`,每行多出可编辑的权重输入框(5 / 3 / 2 / 1):

routes weighted

**`weighted_random` + 会话亲和开启** —— 横幅多出一枚 📌 `Affinity on · by conversation · 30m` 徽章,直接暴露 sticky 的开关状态、scope 和 TTL:

routes affinity

## Summary by CodeRabbit

* **新功能**
  * 在路由列表上方新增“有效路由策略”提示横幅，展示当前生效策略、继承/默认状态、并可跳转至配置页
  * 在加权随机策略下显示权重与会话亲和（sticky）状态，支持 Pin 标识与 TTL 显示
  * 在路由列表中新增内联权重编辑器，允许直接修改并提交路由权重（含取消/回退与输入校验）

* **文档/本地化**
  * 补充中/英文路由策略、权重与亲和相关的本地化文案与提示

### DIFF
--- a/web/src/components/routes/ClientTypeRoutesContent.tsx
+++ b/web/src/components/routes/ClientTypeRoutesContent.tsx
@@ -5,7 +5,8 @@
 
 import { useState, useMemo, useRef } from 'react';
 import { useTranslation } from 'react-i18next';
-import { Plus, RefreshCw, Zap } from 'lucide-react';
+import { Link } from 'react-router-dom';
+import { Plus, RefreshCw, Zap, Workflow, Settings2 } from 'lucide-react';
 import {
   DndContext,
   closestCenter,
@@ -32,19 +33,21 @@ import {
   useUpdateRoutePositions,
   useProviderStats,
   useProxyRequestUpdates,
+  useRoutingStrategies,
   routeKeys,
 } from '@/hooks/queries';
 import { useQueryClient } from '@tanstack/react-query';
 import { useStreamingRequests } from '@/hooks/use-streaming';
 import { getClientName, getClientColor } from '@/components/icons/client-icons';
 import { getProviderColor, type ProviderType } from '@/lib/theme';
-import type { ClientType, Provider, ProviderStats } from '@/lib/transport';
+import type { ClientType, Provider, ProviderStats, RoutingStrategyType } from '@/lib/transport';
 import {
   SortableProviderRow,
   ProviderRowContent,
 } from '@/pages/client-routes/components/provider-row';
 import type { ProviderConfigItem } from '@/pages/client-routes/types';
-import { Button } from '../ui';
+import { Button, Badge, buttonVariants } from '../ui';
+import { cn } from '@/lib/utils';
 import { AntigravityQuotasProvider } from '@/contexts/antigravity-quotas-context';
 import { CooldownsProvider } from '@/contexts/cooldowns-context';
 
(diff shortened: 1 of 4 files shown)


awsldev:
Review note from checking the inline weight editor behavior:

The new `Esc` cancel path in `RouteWeightControlBase` can still commit the in-progress value because it calls `blur()` after `setValue(String(route.weight ?? 1))`, and the input's `onBlur` always runs `commit()`.

Current flow in `provider-row.tsx`:

```tsx
} else if (e.key === 'Escape') {
  setEditing(false);
  setValue(String(route.weight ?? 1));
  (e.target as HTMLInputElement).blur();
}
```

`setValue(...)` is not guaranteed to be reflected before the blur handler reads `value`, so pressing Esc after typing a changed weight can still send `updateRoute.mutate(...)` from `commit()` instead of cancelling.

Suggested fix: add an explicit cancel guard/ref for the next blur, or split `cancel()` from `commit()` so Escape resets local state without invoking the blur commit path. That keeps the intended blur/Enter commit behavior while making Esc a true cancel.

Bowl42:
Good catch — confirmed and fixed in 857c2bb.

You're right: `Escape` called `setValue(原值)` then `blur()`, but `onBlur={commit}` runs synchronously within the keydown handler and reads the **pre-reset** `value` from its render closure, so a changed weight was still committed. `setValue` hadn't taken effect yet.

Rather than thread the reset through async state, I added a `cancelRef` (a ref, since blur fires before any state update from this render is visible to `commit`). Escape sets `cancelRef.current = true` then blurs; `commit` sees the flag, restores the original value, and returns early without mutating. Blur/Enter commit behavior is unchanged.

```tsx
const cancelRef = useRef(false);

const commit = () => {
  setEditing(false);
  if (cancelRef.current) {
    cancelRef.current = false;
    setValue(String(route.weight ?? 1));
    return;
  }
  // ...normalize + mutate only when changed
};

// Escape:
cancelRef.current = true;
(e.target as HTMLInputElement).blur();
```

Also applied the fullwidth-parens nitpick on `weightTooltip`. `tsc` clean.

awsl233777 (approved):
Approved after re-checking the current head. Checks are green, CodeRabbit is successful, and the Esc-cancel issue raised in review was fixed in 857c2bb with a cancelRef guard so blur no longer commits cancelled edits.

PR #608, opened 2026-06-12
fix: hide excluded provider secrets

awsl233777:
## Summary
- Treat `excludeFromExport=true` providers as write-only for saved secrets in admin and self-service provider APIs.
- Preserve existing provider secrets when edit requests submit an empty secret, so editing other fields does not clear hidden credentials.
- Keep existing write-only mode locked on provider update, so a direct API `PUT` cannot turn `excludeFromExport` back off and reveal preserved secrets.
- Hide the API key on the custom provider edit page when provider secrets are write-only, remove the reveal button, and lock the exclude-from-export switch after it is enabled.
- Redact custom-provider request event headers before logging/request-detail delivery, including `Authorization`, `X-Api-Key`, `X-Goog-Api-Key`, AWS session tokens, and cookies.
- Update zh/en copy and add regression tests for excluded provider reads, updates, exports, and event-header redaction.

## Validation
- `go test ./internal/handler ./internal/service`
- `go test ./internal/adapter/provider/custom ./internal/handler ./internal/service`
- `go test ./...`
- `pnpm typecheck`
- `pnpm lint`
- `pnpm build` (passed with existing Vite chunk-size warning)
- Manual E2E against local backend `:9880` and Vite frontend `:3000` using temporary DBs:
  - created a custom provider with `excludeFromExport=true`
  - confirmed `/api/providers` and `/api/providers/:id` redact the saved API key
  - confirmed edit page does not reveal the saved API key and has no reveal button near the API key input
  - saved an edit with empty API key and confirmed the DB still preserves the original test secret
  - attempted a direct malicious `PUT /api/providers/:id` with `excludeFromExport=false`; response and follow-up read kept `excludeFromExport=true` and redacted the API key
  - confirmed `/api/providers/export`, `/api/admin/providers/export`, and `/api/admin/backup/export` do not include the excluded provider
  - scanned edit page text, HTML, localStorage, and sessionStorage for the test secret; no leak found

## Summary by CodeRabbit

* **新功能**
  * 新增“排除从导出”选项与“仅写入”密钥模式：编辑页不回显已保存密钥，输入新值可轮换；克隆/保存流程考虑该模式。

* **隐私/改进**
  * 列表/详情/API 写入响应按角色与导出排除策略条件性隐藏敏感字段。
  * 事件/日志中的请求头敏感字段将被脱敏，不再明文记录凭证。

* **测试**
  * 新增多项测试覆盖密钥隐藏、更新保留与头部脱敏行为。

* **本地化**
  * 更新中/英文文案，说明密钥隐藏与写入模式操作提示。

### DIFF
--- a/internal/adapter/provider/custom/adapter.go
+++ b/internal/adapter/provider/custom/adapter.go
@@ -229,7 +229,7 @@ func (a *CustomAdapter) Execute(c *flow.Ctx, provider *domain.Provider) error {
 		eventChan.SendRequestInfo(&domain.RequestInfo{
 			Method:  upstreamReq.Method,
 			URL:     upstreamURL,
-			Headers: flattenHeaders(upstreamReq.Header),
+			Headers: sanitizeHeadersForEvent(upstreamReq.Header),
 			Body:    string(requestBody),
 		})
 	}
@@ -374,7 +374,7 @@ func (a *CustomAdapter) retryWithFixer(
 			eventChan.SendRequestInfo(&domain.RequestInfo{
 				Method:  retryReq.Method,
 				URL:     retryReq.URL.String(),
-				Headers: flattenHeaders(retryReq.Header),
+				Headers: sanitizeHeadersForEvent(retryReq.Header),
 				Body:    string(fixedBody),
 			})
 		}
@@ -993,6 +993,36 @@ func flattenHeaders(h http.Header) map[string]string {
 		}
 	}
 	return result
+}
+
+// sanitizeHeadersForEvent returns a header map safe for request event logging,
+// redacting upstream credentials that may be injected from provider config.
+func sanitizeHeadersForEvent(h http.Header) map[string]string {
+	result := flattenHeaders(h)
+	for key := range result {
+		if isSensitiveEventHeader(key) {
+			result[key] = "[REDACTED]"
+		}
+	}
+	return result
+}
+
+func isSensitiveEventHeader(key string) bool {
+	switch strings.ToLower(key) {
+	case "authorization",
+		"proxy-authorization",
+		"x-api-key",
+		"x-goog-api-key",
+		"api-key",
+		"anthropic-api-key",
+		"openai-api-key",
+		"x-amz-security-token",
+		"cookie",
+		"set-cookie":
+		return true
+	default:
+		return false
+	}
 }
 
 // Headers to filter out - only privacy/proxy related, NOT application headers like anthropic-version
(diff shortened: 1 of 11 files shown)


awsldev:
Added a follow-up hardening pass after testing the excluded-provider flow from user-accessible paths.

What changed in the second commit:
- Locked existing `excludeFromExport=true` providers at the service layer so a direct API `PUT` cannot switch the flag back to `false` and reveal preserved secrets.
- Redacted custom-provider request event headers before they are sent to request detail/event logging paths (`Authorization`, `X-Api-Key`, `X-Goog-Api-Key`, AWS session token, cookies, etc.).
- Added regression coverage for both the locked write-only mode and custom-provider event-header redaction.

Validation run locally:
- `go test ./internal/adapter/provider/custom ./internal/handler ./internal/service`
- `go test ./...`
- `pnpm typecheck`
- `pnpm lint`
- `pnpm build` (passed with the existing Vite chunk-size warning)
- Manual attack-style E2E on a temporary local DB: created an excluded custom provider, checked list/detail/edit DOM/storage/export/admin export/backup export, attempted malicious `PUT excludeFromExport=false`, and found no test-secret leak after the hardening.

@coderabbitai re-review

awsldev:
Handled the CodeRabbit clone-flow finding.

What changed:
- Hidden-secret providers now require a newly entered API key before cloning.
- The clone path no longer falls back to a redacted/empty stored key for write-only providers.
- Added zh/en validation copy and an inline error near the API key field.

Validation:
- `pnpm typecheck`
- `pnpm lint`
- `pnpm build` (passed with the existing Vite chunk-size warning)
- Manual UI E2E on a temporary local DB: created an `excludeFromExport=true` provider, opened edit, clicked Clone without entering a new key, confirmed the validation message appears and no clone provider is created.

@coderabbitai re-review

awsldev:
Handled the remaining CodeRabbit provider-type preservation finding.

What changed:
- `preserveEmptyProviderSecrets` now only preserves secrets for the effective provider type.
- If an update explicitly changes provider type, old-type config/secrets are not copied into the new provider config.
- Added a regression test that updates a custom provider containing `old-custom-secret` into a Bedrock provider and verifies the stored/response config no longer contains the old custom config.

Validation:
- `go test ./internal/handler ./internal/service`
- `go test ./...`
- `pnpm typecheck`
- `pnpm lint`
- `pnpm build` (passed with the existing Vite chunk-size warning)

@coderabbitai re-review

PR #613, opened 2026-06-17
feat: hide no-export provider URLs

awsldev:
## Summary
- add write-only handling for providers marked `excludeFromExport`, hiding saved base URLs, client-specific URLs, and API keys from admin responses
- preserve the hidden provider URL/key when editing with blank values, and keep `excludeFromExport` irreversible for existing hidden providers
- move the edit-page switch to clone-only options while keeping the create-page switch, with updated Chinese/English copy

## Verification
- `go test ./...`
- `pnpm --dir web typecheck`
- `pnpm --dir web test`
- `pnpm --dir web build`
- `pnpm --dir web lint`

## UI evidence
- Create page shows the no-export switch
- Edit page hides excluded provider endpoint/key and keeps clone-only no-export option

## 发布说明

* **Bug Fixes**
  * 增强了被排除导出的提供商配置的数据保护，确保相关 URL 和敏感信息不会在编辑或克隆时意外泄露。

* **Documentation**
  * 更新了中英文界面文案，新增关于排除导出、写入隐藏和克隆选项的提示说明。

### DIFF
--- a/internal/handler/self_service.go
+++ b/internal/handler/self_service.go
@@ -977,6 +977,10 @@ func sanitizeProvider(provider *domain.Provider) *domain.Provider {
 	if config.Custom != nil {
 		custom := *config.Custom
 		custom.APIKey = ""
+		if provider.ExcludeFromExport {
+			custom.BaseURL = ""
+			custom.ClientBaseURL = nil
+		}
 		if custom.Disguise != nil {
 			disguise := *custom.Disguise
 			if disguise.ClaudeCode != nil {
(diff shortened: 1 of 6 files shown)


awsldev:
Implemented the no-export provider visibility changes in this PR.

What changed:
- Existing `excludeFromExport` providers now behave as write-only for `baseURL`, `clientBaseURL`, and API key in admin/self-service responses.
- Editing an excluded provider with blank URL/key preserves the stored hidden values, and attempts to submit `excludeFromExport: false` keep hidden mode enabled.
- The edit form no longer exposes an original-provider no-export toggle; the switch is now clone-only on the edit page, while create still supports setting it.
- Added regression coverage for hidden URL/client URL sanitization and preservation.

Verification run locally:
- `go test ./...`
- `pnpm --dir web typecheck`
- `pnpm --dir web test`
- `pnpm --dir web build`
- `pnpm --dir web lint`

@coderabbitai review

PR #615, opened 2026-06-17
fix(codex): scope in-stream error events, don't freeze the whole provider

Bowl42:
## Summary

A request to an unsupported model on a ChatGPT-account-backed Codex provider freezes the entire provider for the cooldown window, even though the provider is healthy. Observed on a live setup where `gpt-5.5-codex` is rejected by ChatGPT-account Codex and the upstream returns:

```
{"detail":"The 'gpt-5.5-codex' model is not supported when using Codex with a ChatGPT account."}
```

For ~5 minutes afterward, the affected provider returns "no routes available" for *every* model, not just the rejected one. With two providers both on ChatGPT accounts (a common dual-account setup), both freeze simultaneously and all codex traffic dies.

## Root cause

The Codex Responses API returns 200 OK, emits a structured error event (e.g. `response.failed` with `code=model_not_supported`), and closes the stream. `handleStreamResponse` only watches for `response.completed` and hits the catch-all "stream closed before response.completed" path at `internal/adapter/provider/codex/adapter.go:525-528`, which sets `ScopeProvider + CooldownReasonNetworkError` unconditionally.

The status-code → scope mapping for non-stream responses already has nuanced logic (404 + `model` keyword → `ScopeModel`; 503 + `overloaded`/`model` → `ScopeModel`). The streaming-error path didn't.

## Change

Track the last structured error event observed during the SSE loop. On stream-close-without-completed, `classifyCodexStreamError` maps known signals to narrower scopes; anything not classified keeps the conservative `ScopeProvider + NetworkError` default so genuine outages still trip the wider cooldown.

| Signal | Scope | Reason |
|---|---|---|
| `model_not_supported` / `model_not_found` / "model is not supported" / "model not available" / "no access to the model" / "does not have access to model" / "unknown model" | `ScopeModel` | `ModelUnavailable` |
| `invalid_api_key` / "unauthorized" / "authentication failed" | `ScopeKey` | `AuthFailure` |
| `rate_limit_exceeded` / "rate limit" | `ScopeKey` | `RateLimitExceeded` |
| `insufficient_quota` / "quota" | `ScopeKey` | `QuotaExhausted` |
| unrecognized | `ScopeProvider` | `NetworkError` (unchanged) |

When a captured event narrows to `ScopeModel`, the model name is taken from the event itself (`response.failed.response.model` or `error.model`) since `extractModelFromSSELine` only reads top-level `model` and would miss the nested Responses API shape.

## Tests added

- `TestParseCodexStreamErrorLine` — three wire shapes (`response.failed` / `response.error`, generic `error`, FastAPI-style `detail`) plus non-error events / `[DONE]` / non-data / invalid JSON.
- `TestClassifyCodexStreamError` — table test covering each classification branch including "unrecognized stays unclassified".
- `TestHandleStreamResponseScopesModelOnInStreamModelError` — drives the real wire shape end-to-end and asserts `ScopeModel + ModelUnavailable + Model = gpt-5.5-codex`.
- `TestHandleStreamResponseKeepsProviderScopeWithoutErrorEvent` — guards the conservative fallback for unknown stream truncation.

## Test plan

- [x] `go build ./...`
- [x] `go vet ./internal/adapter/provider/codex/...`
- [x] `go test -count=1 ./internal/adapter/provider/codex/... ./internal/executor/... ./internal/cooldown/...` — all green
- [ ] Live verification: stream a request for an unsupported model, confirm only that model is cooled down (other models on the same provider keep serving) and that the cooldown row carries the model name + `model_unavailable` reason.

## 改进

* **Bug 修复**
  * 增强了流式 API 错误检测机制，提升错误分类的精准度，改进故障诊断能力

* **测试**
  * 新增测试用例覆盖错误处理的多种场景

### DIFF
--- a/internal/adapter/provider/codex/adapter.go
+++ b/internal/adapter/provider/codex/adapter.go
@@ -534,14 +534,26 @@ func (a *CodexAdapter) handleStreamResponse(c *flow.Ctx, resp *http.Response) er
 			// frozen along with the failing one. Anything we can't classify
 			// confidently keeps the conservative ScopeProvider fallback.
 			if scope, reason, ok := classifyCodexStreamError(lastStreamErr); ok {
-				proxyErr.Scope = scope
-				proxyErr.Reason = reason
 				if scope == domain.ScopeModel {
 					proxyErr.Model = lastStreamErr.model
 					if proxyErr.Model == "" {
 						proxyErr.Model = model
 					}
+					if proxyErr.Model == "" {
+						proxyErr.Model = flow.GetMappedModel(c)
+					}
+					// Without a model to attribute, ScopeModel would collapse
+					// to a (provider,"","") cooldown key — i.e. provider-wide —
+					// defeating the point of refining the scope. Fall back to
+					// the conservative ScopeProvider in that case.
+					if proxyErr.Model == "" {
+						proxyErr.Scope = domain.ScopeProvider
+						proxyErr.Reason = domain.CooldownReasonNetworkError
+						return proxyErr
+					}
 				}
+				proxyErr.Scope = scope
+				proxyErr.Reason = reason
 			} else {
 				proxyErr.Scope = domain.ScopeProvider
 				proxyErr.Reason = domain.CooldownReasonNetworkError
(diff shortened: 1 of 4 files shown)


awsldev (approved):
Approved after green CI, targeted codex adapter tests, CodeRabbit timeout on latest head, and manual review confirming the previous classification blocker was fixed.

PR #628, opened 2026-06-25
fix(providers): constrain bulk import dialog height

awsldev:
## Summary
- Constrain the bulk custom provider import dialog to the viewport height.
- Keep the dialog header/footer visible while the command input and preview area scroll in the middle.
- Stop the bulk command textarea from auto-growing with pasted content by using fixed field sizing.

## Validation
- Reproduced the old layout issue with ~42 `provider add` lines: dialog height grew to about 5696px in a 1280×800 viewport.
- Verified fixed desktop layout at 1280×800: dialog stays within viewport and footer remains visible.
- Verified fixed narrow layout at 390×700: footer remains fully visible.
- `pnpm test src/pages/providers/utils/bulk-custom-provider-import.test.ts` — 6/6 passed
- `pnpm test` — 29/29 passed
- `pnpm typecheck` — passed
- `pnpm lint` — passed
- `pnpm build` — passed; Vite reported the existing large chunk warning only.


awsldev:
Closing this PR because it was opened from an old feature branch that carries stale bulk-import commits and conflicts with current main. Reopening from a clean branch containing only the dialog height fix.

PR #643, opened 2026-06-30
fix(ui): keep token nav visible and clarify cooldown copy

awsldev:
## Summary
- keep the API Tokens sidebar entry visible for admin users even when admin auth is not enabled
- keep the entry guarded as admin-only and cover that contract with a regression test
- clarify the zh/en provider copy for `disableErrorCooldown`, including retry-policy behavior and existing cooldown handling

## Why
Packaged builds include the API Tokens page, but the sidebar entry was marked `authOnly`. In the default unauthenticated admin setup, that hides the entry even though the page is used to enable API token authentication.

The error-cooldown setting copy also needed to match the current behavior more precisely: enabling the switch prevents new automatic cooldowns from upstream errors and related request errors continue according to the configured retry policy, while existing cooldowns still need to expire or be cleared manually.

## Verification
- `pnpm exec prettier --check src/locales/zh.json src/locales/en.json`
- parsed `web/src/locales/zh.json` and `web/src/locales/en.json` as JSON
- GitHub CI and CodeRabbit are rerunning after the latest push

### DIFF
--- a/web/src/components/layout/app-sidebar/sidebar-config.test.ts
+++ b/web/src/components/layout/app-sidebar/sidebar-config.test.ts
@@ -1,0 +1,20 @@
+import { describe, expect, it } from 'vitest';
+import { sidebarConfig } from './sidebar-config';
+import type { MenuItem } from '@/types/sidebar';
+
+function findItem(key: string): MenuItem | undefined {
+  return sidebarConfig.sections.flatMap((section) => section.items).find((item) => item.key === key);
+}
+
+describe('sidebarConfig', () => {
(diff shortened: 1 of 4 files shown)


awsldev:
@coderabbitai review

Please re-review this small UI fix. Focus on whether keeping the API Tokens sidebar entry visible without `authOnly` preserves the intended admin-only access model and whether the regression test covers the packaged/default unauthenticated admin setup.

Validated locally:
- `pnpm exec vitest run src/components/layout/app-sidebar/sidebar-config.test.ts`
- `pnpm run typecheck`
- `pnpm run lint`
- `pnpm run build`

PR #663, opened 2026-07-11
feat(router): clear 400 when a requested model is not in any provider's allowlist

Bowl42:
## What & why

A provider can restrict which models it serves via its **`SupportModels`** allowlist (wildcard patterns; empty = allow all). When a client requests a model that matches **no** provider's allowlist, the router dropped every candidate and returned the generic `ErrNoAvailableProviders` → **HTTP 503 "no available provider"**.

That's misleading: 503 reads like a transient upstream outage, so clients back off and retry. But this is a **client-side request error** — the model is disallowed and retrying will never succeed. The caller gets no signal about the real cause.

## Change

- **`router.Match`** now records *why* candidates were dropped. When the candidate set is empty **purely** because of `SupportModels` rejections — and no route was skipped for a transient reason (cooldown) that might have masked a provider which *does* serve the model — it returns a new **`ErrModelNotSupported`** sentinel.
- **route-match middleware** maps `ErrModelNotSupported` to a **non-retryable HTTP 400**, `code: "model_not_supported"`, with a message naming the model (`model "gpt-4o" is not supported by any configured provider`). Scope is `ScopeRequest` (no cooldown, no cross-provider retry).
- **Unchanged:** no-routes and all-providers-in-cooldown still return 503. The transient-skip guard prevents mislabeling a cooled-down-but-supported provider as "unsupported".

## Behavior

| Scenario | Before | After |
|---|---|---|
| Model not in any provider's `SupportModels` | 503 `no_available_provider` | **400 `model_not_supported`** (names the model, non-retryable) |
| No routes for the client type | 503 | 503 (unchanged) |
| All providers in cooldown | 503 | 503 (unchanged) |
| Cooldown hides a provider that *would* serve the model | 503 | 503 (unchanged — not mislabeled) |

## Tests

- Updated the existing `TestProxyMatchedRouteWithUnsupportedProviderModelRecordsRejectedRequest` e2e test to the new 400 / `model_not_supported` contract (it previously pinned the old 503).
- Parameterized the shared `assertRouteMatchRejectionRecorded` helper by expected status code; the three no-routes / cooldown callers keep asserting 503.
- Full `internal/...` + `tests/e2e` suites green; `go vet` + `gofmt` clean.

## Summary by CodeRabbit

* **错误处理**
  * 当请求模型不受任何已配置提供方支持时，现返回“模型不支持”错误，并以 HTTP 400 响应。
  * 错误信息包含请求的模型名称及“不被配置提供方支持”的原因。
  * 路由匹配失败仍保留对临时/冷却等场景的区分，避免错误码被错误归类为模型不支持。
* **测试**
  * 更新了端到端代理集成测试的拒绝断言，增加对状态码与错误消息片段的校验。

### DIFF
--- a/internal/domain/errors.go
+++ b/internal/domain/errors.go
@@ -14,17 +14,22 @@ var (
 	ErrInvalidState         = errors.New("invalid state")
 	ErrNoRoutes             = errors.New("no routes available")
 	ErrNoAvailableProviders = errors.New("no available providers")
-	ErrAllRoutesFailed      = errors.New("all routes failed")
-	ErrFirstByteTimeout     = errors.New("first byte timeout")
-	ErrStreamIdleTimeout    = errors.New("stream idle timeout")
-	ErrUpstreamError        = errors.New("upstream error")
-	ErrFormatConversion     = errors.New("format conversion error")
-	ErrUnsupportedFormat    = errors.New("unsupported format")
-	ErrInviteCodeRequired   = errors.New("invite code required")
-	ErrInviteCodeInvalid    = errors.New("invite code invalid")
-	ErrInviteCodeExpired    = errors.New("invite code expired")
-	ErrInviteCodeExhausted  = errors.New("invite code exhausted")
-	ErrInviteCodeDisabled   = errors.New("invite code disabled")
+	// ErrModelNotSupported means routes/providers existed for the client type but
+	// every candidate was rejected because the requested model is not in the
+	// provider's SupportModels allowlist — a client-side request error (the model
+	// is disallowed), distinct from the transient ErrNoAvailableProviders.
+	ErrModelNotSupported   = errors.New("model not supported by any provider")
+	ErrAllRoutesFailed     = errors.New("all routes failed")
+	ErrFirstByteTimeout    = errors.New("first byte timeout")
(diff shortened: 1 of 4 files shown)


awsldev:
Handled the CodeRabbit test coverage suggestion.

Changes:
- Added `TestProxyMatchedRouteWithUnsupportedModelAndCooldownProviderRecords503`.
- Covers the mixed case where one route is rejected by model allowlist while another provider that supports the requested model is currently in cooldown.
- Verifies the router keeps the generic `503 no_available_provider` path instead of misclassifying the request as unsupported model `400`.

Validation:
- `go test ./tests/e2e -run 'TestProxyMatchedRouteWithUnsupported(ModelAndCooldownProviderRecords503|ProviderModelRecordsRejectedRequest)|TestProxyMatchedRouteWithCooldownProviderRecordsRejectedRequest'`

@coderabbitai review

awsldev (approved):
Reviewed the final diff and checks. The model-allowlist path now returns 400 only when all candidates are rejected by SupportModels, while cooldown/transient skips still preserve the 503 no_available_provider behavior. The added e2e covers the mixed cooldown + model rejection case. Checks are green.

PR #670, opened 2026-07-14
fix: preserve protocol-native user agents

luodaoyi:
## Summary

- preserve the inbound `User-Agent` exactly for protocol-native Codex, Claude, and Gemini requests
- replace the source identity with the target protocol's standard `User-Agent` only after protocol conversion
- stop injecting stale hard-coded Codex `Version` values; preserve an explicit native Codex version and discard source-protocol versions during conversion
- pass only `User-Agent` and `Version` into the embedded CLIProxyAPI Codex executor context

## Root cause

The adapters selected headers from the target provider without considering whether the request protocol had changed. Missing or unrecognized user agents fell back to hard-coded identities, and Codex paths injected fixed `Version` values (`0.21.0` or `0.98.0`). Current Codex clients could therefore be rejected as outdated even when the installed CLI was current.

The embedded CLIProxyAPI adapter also did not expose the inbound Codex identity headers through the Gin request context consumed by its executor.

## Behavior after this change

- native request: retain the client's `User-Agent`, including a deliberately blank value
- converted request: discard the source identity and use the target protocol default
- native Codex request: retain an explicit client `Version`; do not invent one when absent
- converted-to-Codex request: remove any source-protocol `Version`
- embedded CPA request: expose only `User-Agent` and `Version`, avoiding credentials and unrelated header-driven behavior

## Relationship to #669

This PR is stacked on #669 while the CLIProxyAPI dependency update is under review. It does not add or change dependency metadata. After #669 merges, the base will be changed to `main`.

## Validation

- `go test -count=1 ./internal/flow ./internal/adapter/provider/custom ./internal/adapter/provider/codex ./internal/adapter/provider/claude ./internal/adapter/provider/cliproxyapi_codex ./internal/adapter/provider/cliproxyapi_antigravity`
- `go vet ./cmd/... ./internal/...`
- `go build -o .codex-maxx-ci-build.exe ./cmd/maxx`
- `pnpm run lint`, `pnpm run typecheck`, and `pnpm run build` in `web/`
- `git diff --check`

If this behavior regresses, inspect `original_client_type`, `client_type`, and the final upstream `User-Agent` and `Version` values to identify whether the fault is in dispatch, protocol conversion, adapter header selection, or the embedded CPA executor boundary.

### DIFF
--- a/internal/adapter/provider/claude/adapter.go
+++ b/internal/adapter/provider/claude/adapter.go
@@ -124,6 +124,7 @@ func (a *ClaudeAdapter) Execute(c *flow.Ctx, provider *domain.Provider) error {
 
 	// Apply headers
 	a.applyClaudeHeaders(upstreamReq, request, accessToken, clientWantsStream, extraBetas)
+	upstreamReq.Header["User-Agent"] = []string{flow.ResolveUpstreamUserAgent(c, ClaudeUserAgent)}
 
 	// Send request info via EventChannel
 	if eventChan := flow.GetEventChan(c); eventChan != nil {
@@ -171,6 +172,7 @@ func (a *ClaudeAdapter) Execute(c *flow.Ctx, provider *domain.Provider) error {
 			return proxyErr
 		}
 		a.applyClaudeHeaders(upstreamReq, request, accessToken, clientWantsStream, extraBetas)
+		upstreamReq.Header["User-Agent"] = []string{flow.ResolveUpstreamUserAgent(c, ClaudeUserAgent)}
 
 		resp, err = a.httpClient.Do(upstreamReq)
 		if err != nil {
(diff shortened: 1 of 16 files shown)


awsldev (approved):

awsldev:
Review note from checking the current UA conversion change:

The main adapter paths look directionally right: native protocol requests preserve the inbound `User-Agent`, and normal protocol conversion paths call `ResolveUpstreamUserAgent(...)`, so they use the target protocol default instead of leaking the source UA.

One boundary still needs tightening before I would call the described requirement complete: the embedded CLIProxyAPI Codex path deletes both `User-Agent` and `Version` during protocol conversion, but the PR only tests the intermediate executor headers/context, not the final upstream header emitted by CLIProxyAPI.

Evidence in this PR:

```go
if originalClientType != "" && targetClientType != "" && originalClientType != targetClientType {
    headers.Del("User-Agent")
    headers.Del("Version")
}
```

That can be fine for OAuth-backed CPA because CLIProxyAPI can fill its Codex default, but API-key CPA header handling may keep `User-Agent` empty when no gin header/default is supplied. So the current test coverage does not prove that converted-to-CPA requests always get the target Codex UA.

Suggested fix: add a regression test at the embedded CPA boundary that covers a converted request all the way through the CPA header selection path, including the API-key/auth mode if supported. The expected result should be: converted request does not preserve source UA, and the final upstream Codex request has the target Codex UA rather than an empty UA.

Until that is covered or fixed, I would not mark the UA-conversion requirement as fully complete.

PR #682, opened 2026-07-15
fix(codex): sanitize WebSocket HTTP fallback

luodaoyi:
## Summary

- preserve the official Responses WebSocket v2 `generate: false` prewarm field on native Codex upstream WebSocket frames
- strip the WebSocket-only `generate` field from MAXX's normalized HTTP/SSE fallback body
- log why a native upstream WebSocket session falls back, so Docker/proxy upgrade failures can be diagnosed

## Root cause

MAXX keeps the original downstream WebSocket frame in request context for the native Codex provider path, while also constructing an HTTP Responses body for fallback. The fallback body still contained `generate`, so if the upstream WebSocket upgrade could not be established, MAXX POSTed that WebSocket-only field to CPA. CPA correctly rejected it as an unsupported ordinary HTTP Responses parameter.

This change keeps the two transport contracts separate: native WebSocket gets `generate: false`; HTTP fallback does not.

## Verification

- `go test ./internal/adapter/provider/codex ./internal/handler ./internal/core -count=1`
- `go vet ./internal/adapter/provider/codex ./internal/handler ./internal/core`
- `go test ./... -count=1` (all relevant packages and e2e pass; two pre-existing Windows-only failures remain in `cmd/maxx-cli/internal/cfg`: Unix mode 0600 expectation and legacy temp-path migration)
- `git diff --check`

The existing pprof behavior remains unchanged and verified: listener address is built with `0.0.0.0`, while the settings link uses the browser's current hostname.

## Summary by CodeRabbit

* **改进**
  * 优化 WebSocket 请求转发，避免内部参数泄漏到后端请求中。
  * 在 WebSocket 不可用、升级失败或写入失败时，自动回退至 HTTP/SSE 通道，提升连接稳定性。
  * 增强回退场景的日志记录，便于排查连接问题。

### DIFF
--- a/internal/adapter/provider/codex/websocket.go
+++ b/internal/adapter/provider/codex/websocket.go
@@ -6,6 +6,7 @@ import (
 	"errors"
 	"fmt"
 	"io"
+	"log"
 	"net"
 	"net/http"
 	"net/url"
@@ -189,6 +190,7 @@ func (a *CodexAdapter) executeResponsesWebSocket(c *flow.Ctx, provider *domain.P
 	if session == nil && previousResponseID != "" {
 		// A response ID is tied to the provider/session that produced it. If route
 		// selection moved elsewhere, use the handler's full-transcript HTTP fallback.
+		log.Printf("[Codex] Responses WebSocket session unavailable for provider=%d; falling back to HTTP/SSE", provider.ID)
 		return false, nil
 	}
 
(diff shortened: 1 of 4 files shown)


ymkiux (approved):

awsldev (approved):
Approved for merge: required checks are passing and PR is mergeable.

PR #710, opened 2026-07-18
feat(openrouter): image aspect ratio across protocols + generated-image passthrough

Bowl42:
## Problem

Image generation config (aspect ratio / size) was lost when a request crossed protocols to reach an OpenAI-speaking upstream (e.g. OpenRouter), and generated images were dropped when converting the response back to a Gemini client. Each API protocol expresses image sizing its own standard way — OpenAI uses a pixel `size`, Gemini uses `generationConfig.imageConfig` — and these weren't preserved across conversion.

## What this does

Each protocol sends its own standard; conversion is handled transparently so the same client request works regardless of which upstream serves it.

**Request side** (`feat(openrouter): preserve image aspect ratio across protocols`)
- Converter (gemini→openai): carry `imageConfig`/`responseModalities` into OpenAI `image_config`/`modalities` so a Gemini client's aspect ratio survives the forced Gemini→OpenAI conversion. Generic — benefits any OpenAI-speaking upstream.
- OpenRouter provider: endpoint-aware normalization before forwarding — chat honors `image_config.aspect_ratio` (+`modalities`), the images endpoint honors pixel `size`; fill in whichever the endpoint needs from what the client sent.

**Response side** (`fix(converter): carry generated images openai->gemini response`)
- Map the non-standard `message.images[]` array (how OpenRouter/Gemini image models return generated images) to Gemini `parts[].inlineData`, in both the non-stream and streaming response converters — so a Gemini client actually receives the image.

## Testing

- Unit tests for both converters and the provider normalization.
- e2e integration tests through the full proxy pipeline against a mock upstream (CI-safe): OpenAI chat size→aspect, images aspect→size, and a full Gemini-client→OpenRouter round trip.
- Verified end-to-end against live OpenRouter (decoding returned image dimensions): OpenAI client `size=1536x1024` → 3:2, Gemini client `aspectRatio=9:16` → 9:16 with the image returned. Also verified the same request fails over correctly across upstreams.

## Summary by CodeRabbit

- **新功能**
  - 支持 Gemini 图像生成参数转换为 OpenAI/OpenRouter 格式，包括图像尺寸、长宽比及响应模态。
  - 自动兼容不同接口的图像尺寸与长宽比参数。
  - 支持将 OpenAI/OpenRouter 返回的生成图片转换为 Gemini 可识别的图片内容，覆盖普通和流式响应。

- **Bug 修复**
  - 改善图像生成请求在不同接口间的参数兼容性，确保图像模态信息正确传递。

### DIFF
--- a/internal/adapter/provider/openrouter/adapter.go
+++ b/internal/adapter/provider/openrouter/adapter.go
@@ -92,7 +92,15 @@ func (a *Adapter) SupportedClientTypes() []domain.ClientType {
 }
 
 // Execute delegates to the custom core, passing the synthesized provider so the
-// custom config drives base URL and auth.
+// custom config drives base URL and auth. Before delegating it normalizes any
+// image sizing into the form the target OpenRouter endpoint honors (see
+// normalizeImageConfigBody), so each protocol can express sizing its own standard
+// way and still reach the upstream image model correctly.
 func (a *Adapter) Execute(c *flow.Ctx, _ *domain.Provider) error {
+	if body := flow.GetRequestBody(c); len(body) > 0 {
+		if nb := normalizeImageConfigBody(body, flow.GetRequestURI(c)); len(nb) > 0 {
+			c.Set(flow.KeyRequestBody, nb)
+		}
+	}
 	return a.inner.Execute(c, a.synth)
 }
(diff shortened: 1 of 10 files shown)


awsldev (approved):
Approved after checking mergeability, required checks, and targeted regression coverage for OpenRouter image config and Gemini/OpenAI image conversion.

PR #746, opened 2026-07-23
fix(executor): keep smart mapping retry on disabled switch provider

awsldev:
## Summary
- prevent DisableErrorCooldown + SmartMappingRetry exhaustion from falling through to the next provider route
- cycle mapped-model candidates on the same provider when provider switching is disabled
- add regression coverage proving the second provider is not called while smart mapping polling retries continue on the original provider

## Validation
- go test ./internal/executor
- go test ./internal/executor ./internal/router ./internal/handler ./internal/service

CodeRabbit not checked/triggered unless explicitly requested.

### DIFF
--- a/internal/executor/disabled_error_cooldown_test.go
+++ b/internal/executor/disabled_error_cooldown_test.go
@@ -242,6 +242,111 @@ func TestDispatchSmartMappingRetrySwitchesMappedModelAfterLimit(t *testing.T) {
 	}
 }
 
+type noProviderSwitchSmartMappingAdapter struct {
+	models       []string
+	succeedAfter int
+}
+
+func (a *noProviderSwitchSmartMappingAdapter) SupportedClientTypes() []domain.ClientType {
+	return []domain.ClientType{domain.ClientTypeOpenAI}
+}
+
+func (a *noProviderSwitchSmartMappingAdapter) Execute(c *flow.Ctx, _ *domain.Provider) error {
+	model := flow.GetMappedModel(c)
+	a.models = append(a.models, model)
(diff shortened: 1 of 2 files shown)


awsldev:
Validation before admin merge:

- Reviewed the final diff: executor smart-mapping retry exhaustion no longer falls through to the next provider when DisableErrorCooldown is enabled.
- Added regression coverage for the user-visible scenario: two provider routes, DisableErrorCooldown + SmartMappingRetry enabled on the first provider, mapped models cycle `mapped-a -> mapped-b -> mapped-a`, and the second provider is never called.
- Local validation passed:
  - `go test ./internal/executor -run 'TestDispatchDisableErrorCooldownKeepsSmartMappingRetryOnSameProvider|TestDispatchSmartMappingRetrySwitchesMappedModelAfterLimit|TestDispatchDisableErrorCooldownRetriesHTTPErrorBeyondRetryBudget' -count=20`
  - `go test ./internal/executor ./internal/router ./internal/handler ./internal/service ./internal/cooldown ./internal/domain ./internal/converter`
  - `cd web && pnpm typecheck`
  - `cd web && pnpm test` (29 files / 137 tests)
  - `cd web && pnpm build`
  - `cd web && pnpm lint`
  - `go test ./...` after building `web/dist`
- Remote check observed successful for the PR check. The Docker workflow's only job `build-and-push-image` completed successfully; the workflow wrapper remained stale/in_progress at the time of merge gating.
- No failed check results observed.

CodeRabbit was not checked/triggered because this flow did not request CodeRabbit review.

PR #755, opened 2026-07-24
fix(codex): stop WebSocket 1012 replay reconnect loops

luodaoyi:
## Summary

- preserve upstream Responses WebSocket close code and reason in structured attempt errors
- forward safe upstream close codes 1009, 1012, and 1013 before emitting a JSON error frame
- add a 30-second WebSocket-only provider transport cooldown for upstream 1012/1013
- make route matching and handshake preflight use the same WebSocket transport availability state
- keep HTTP/SSE routing available while the WebSocket transport is cooling down
- sanitize close reasons to valid UTF-8 within the 123-byte control-frame limit

## Why

An upstream CPA 1012 Service Restart was converted into a JSON 502 plus downstream 1011. Codex then reconnected, while the handshake preflight still treated the provider as WebSocket-capable. That allowed repeated upgrade/replay cycles and elevated CPU usage.

After this change, the original 1012 reaches the client, the affected provider is temporarily excluded only from WebSocket routing, and the next handshake returns HTTP 426 when no other WebSocket provider is available so Codex can fall back to HTTP/SSE.

## Validation

- `go vet ./cmd/... ./internal/...`
- `go test ./internal/adapter/provider ./internal/adapter/provider/custom ./internal/adapter/provider/codex ./internal/domain ./internal/handler ./internal/router -count=1`
- `go build -o /tmp/maxx-ci-build ./cmd/maxx`
- `pnpm run lint`
- `pnpm run typecheck`
- `pnpm run build`

## Summary by CodeRabbit

* **新功能**
  * 上游 WebSocket 触发“服务重启/稍后重试”时，自动进行短暂传输冷却，并可按需立即清理以恢复可用性。
  * 上游 WebSocket 关闭码与关闭原因会被更准确记录，并在符合条件时转发给客户端。
* **问题修复**
  * 路由选择更智能：冷却期间仅影响 WebSocket 升级路径，不影响其他连接方式。
  * 关闭原因在转发前进行 UTF-8 校验与长度限制，避免异常/超长内容。
* **测试/保障**
  * 新增覆盖冷却、关闭分类与转发行为的测试用例。

### DIFF
--- a/internal/adapter/provider/codex/websocket.go
+++ b/internal/adapter/provider/codex/websocket.go
@@ -16,6 +16,7 @@ import (
 	"sync/atomic"
 	"time"
 
+	provideradapter "github.com/awsl-project/maxx/internal/adapter/provider"
 	"github.com/awsl-project/maxx/internal/domain"
 	"github.com/awsl-project/maxx/internal/flow"
 	"github.com/awsl-project/maxx/internal/usage"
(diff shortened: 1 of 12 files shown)


luodaoyi:
Addressed the CodeRabbit maintainability note in a9c2e9b by centralizing upstream CloseError extraction and 1012/1013 transport-cooldown policy in the provider package. Both Codex and Custom adapters now use the shared classifier, with focused regression tests.

luodaoyi:
@coderabbitai review

PR #810, opened 2026-08-15
fix(router): reconcile provider adapters after cache invalidation

Bowl42:
## Summary
- rebuild router provider adapters after cross-instance provider cache invalidation
- wire provider invalidation to reload provider cache and reconcile runtime adapters
- add a multi-instance integration test for project route visibility and adapter availability across instances

## Tests
- /usr/local/go/bin/go test ./tests/multiinstance
- /usr/local/go/bin/go test ./internal/core ./internal/router ./cmd/maxx
- /usr/local/go/bin/go test ./... (fails only for root package: web/dist embed directory is missing; remaining packages pass)

## Summary by CodeRabbit

* **新功能**
  * 多实例环境下，提供商配置变更可自动同步缓存，并更新运行时适配器映射。
  * 项目、提供商与路由信息可在实例间及时同步，确保请求匹配正确的路由和适配器。
* **错误修复**
  * 修复提供商变更后路由或适配器状态未及时更新的问题。
  * 提升跨实例项目路由创建与访问的一致性和可靠性。
* **测试**
  * 新增多实例场景验证，覆盖项目创建、缓存失效、路由匹配及适配器初始化流程。

### DIFF
--- a/cmd/maxx/main.go
+++ b/cmd/maxx/main.go
@@ -237,6 +237,10 @@ func main() {
 	cachedAPITokenRepo := cached.NewAPITokenRepository(apiTokenRepo)
 	cachedModelMappingRepo := cached.NewModelMappingRepository(modelMappingRepo)
 
+	// Create router before wiring invalidation so provider changes from peers
+	// can reload both the provider cache and the runtime adapter map.
+	r := router.NewRouter(cachedRouteRepo, cachedProviderRepo, cachedRoutingStrategyRepo, cachedRetryConfigRepo, cachedProjectRepo, settingRepo)
+
 	// Wire cross-instance cache invalidation. AttachCachedReposToCoordinator
 	// 是 desktop launcher 也走的同一个 helper,保证两条启动路径行为一致。
 	core.AttachCachedReposToCoordinator(coordCtx, coord, &core.DatabaseRepos{
@@ -248,7 +252,7 @@ func main() {
 		CachedAPITokenRepo:        cachedAPITokenRepo,
 		CachedModelMappingRepo:    cachedModelMappingRepo,
 		CachedSessionRepo:         cachedSessionRepo,
-	})
+	}, r)
 
 	// Load cached data
 	startupStep = time.Now()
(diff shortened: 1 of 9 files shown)


awsldev:
Found one merge-blocking concern in `ReconcileAdapters()`.

The current implementation rebuilds a fresh adapter map for every provider invalidation:

```go
providers := r.providerRepo.GetAll()
next := make(map[uint64]provider.ProviderAdapter, len(providers))
// factory(p) for every provider
r.adapters = next
```

That makes a peer-side provider reload much broader than the changed provider. Any provider update can replace every runtime adapter instance, including unchanged Codex/custom providers. Those adapters own runtime state such as Responses WebSocket continuation sessions; replacing the instance drops that state and can leave existing upstream websocket sessions unreachable until their own timeout/cleanup. It can also disrupt unrelated in-flight/continuation requests even though only one provider changed.

Suggested fix: reconcile incrementally instead of wholesale replacement:

- keep existing adapter instances for providers whose relevant config/type did not change;
- refresh/create only added or changed providers;
- remove only deleted providers, and clear provider-specific websocket cooldown/state for those IDs as appropriate;
- add a regression test that an unrelated provider invalidation does not replace/drop an existing adapter instance/session state.

The current tests cover the happy path for a newly-created provider becoming matchable on another instance, but not the state-preservation case for existing adapters. This is risky in the provider adapter / websocket request path, so I would not treat this as merge-ready without tightening that behavior.

Bowl42:
Addressed the merge-blocking adapter lifecycle concern.

Changes pushed in latest commit:
- `ReconcileAdapters()` now reconciles incrementally using an adapter construction fingerprint instead of replacing the whole adapter map.
- Unchanged providers keep the existing adapter instance, preserving runtime state such as websocket continuation/session stores.
- Added/changed providers get a fresh adapter and have only their own websocket transport cooldown cleared.
- Deleted providers are removed and only their provider-specific websocket cooldown is cleared.
- Added `TestReconcileAdaptersPreservesUnchangedStatefulAdapters` covering adapter identity preservation and cooldown isolation across an unrelated provider invalidation.

Local validation:
- `/usr/local/go/bin/go test ./internal/router -count=1`
- `/usr/local/go/bin/go test ./tests/multiinstance -count=1`
- `/usr/local/go/bin/go test ./internal/core ./internal/executor ./internal/handler ./cmd/maxx`

awsldev (approved):
Approved after re-review: conflict-free, required checks passed, and the adapter reconciliation regression concern is addressed with incremental preservation plus targeted tests.

PR #815, opened 2026-08-18
fix(openrouter): always advertise Codex as a native capability (follow-up to #814)

Bowl42:
Follow-up to #814, addressing CodeRabbit's review finding on that PR.

## The gap #814 left

#814 makes Codex→OpenRouter pass through to `/responses` **only when the provider advertises `codex`** in its supported client types (so `NeedConvert` is false). But:

- `CanonicalSupportedClientTypes("openrouter", …)` defaulted to `[claude, openai]` — **no codex**, and
- the OpenRouter adapter reported the **stored** list verbatim (`a.synth.SupportedClientTypes`).

So a pre-existing / default OpenRouter provider (stored as `[claude, openai]`) still fell into the lossy Chat Completions conversion — **the passthrough silently didn't apply**. Only a provider explicitly configured with `codex` (e.g. a hand-edited one) benefited. CodeRabbit flagged exactly this: *"update CanonicalSupportedClientTypes and the OpenRouter adapter's default supported types so production config includes Codex — don't just modify the test adapter's explicit list."*

## Fix

OpenRouter natively speaks Claude (`/v1/messages`), OpenAI (`/v1/chat/completions`) **and** Codex (`/v1/responses`), so Codex is a first-class native capability — not an add-on.

- `internal/domain/provider_capability.go` — the `openrouter` canonical set now **always includes codex** (merged onto whatever is configured / the default), via a small `ensureClientType` helper.
- `internal/adapter/provider/openrouter/adapter.go` — `SupportedClientTypes()` now returns the **canonical** set instead of the possibly-stale stored list, so existing providers advertise codex **without a data migration**.

A Codex request still only reaches a provider when an explicit Codex route points at it — this just ensures that when it does, it passes through natively instead of being converted.

Everything routing-related reads capabilities through the adapter or `ProviderNativelySupports` (both canonical), so no migration is required for runtime correctness.

## Also (CodeRabbit test nits on #814)

- `httptest.NewRequestWithContext` instead of `NewRequest(...).WithContext(...)`.
- `t.Cleanup` invalidation of the process-global `systemsettingcache` so a cached `"true"` can't leak across tests.

## Tests

- `provider_capability_test.go` — openrouter canonical: empty→`[claude,openai,codex]`; `[claude,openai]`→`+codex`; keeps other configured protocols and adds codex; no duplicate codex.
- `openrouter/adapter_test.go` — `SupportedClientTypes()` advertises codex even when the stored config omits it.
[...]

PR #816, opened 2026-08-18
fix(openrouter): pin Codex Responses passthrough to /v1/responses

Bowl42:
## The bug: Codex on OpenRouter returned 200 but empty

OpenRouter serves the OpenAI **Responses API only at `/api/v1/responses`**. A quick probe with a real key:

| URL | result |
|---|---|
| `https://openrouter.ai/api/responses` | **HTTP 200, `text/html`** — the OpenRouter website SPA (~130KB), not the API |
| `https://openrouter.ai/api/v1/responses` | native Responses API JSON (incl. `custom_tool_call` for code-mode tools) |

maxx accepts Codex at both `/responses` and `/v1/responses`, and the custom core echoes the client's *original* Responses path upstream (`flow.KeyResponsesClientPath = r.URL.RequestURI()`). A Codex client whose `base_url` omits `/v1` therefore reaches maxx at bare `/responses` → forwarded to `openrouter.ai/api/responses` → **HTML page** → Codex parses no output and returns **empty**, while the request logs as **COMPLETED / 200**.

This is the "OpenRouter 200 but all broken" report: on a live instance ~**33% of codex→OpenRouter requests came back empty**, the stored upstream body being 130KB of HTML.

The OpenAI path is unaffected because `/v1/chat/completions` carries its own `/v1`; only the Responses path varies by client `base_url`.

## Fix

In the OpenRouter adapter, pin the Codex Responses passthrough to `/v1/responses` regardless of the client path (OpenRouter-specific knowledge: its API root is `/api/v1`). One line in `Execute`, gated on `clientType == codex`.

## Test

`TestOpenRouterMockCodexResponsesPath` posts to the bare `/responses` and asserts the mock upstream is hit at `/v1/responses`. Confirmed it **fails without the fix** (`upstream path = /responses`) and passes with it.

## Relation to #814 / #815

#814 defaulted codex→OpenRouter to native `/responses` passthrough; #815 made every OpenRouter provider advertise codex. Both correct, but they exposed this path bug — the passthrough was hitting the wrong OpenRouter path. The kill switch `codex_openrouter_bridge_enabled=true` remains a stopgap (reverts to the chat bridge); this PR makes native passthrough actually work so it can stay off.

## Summary by CodeRabbit

- **Bug 修复**
  - 修复 Codex 请求在 OpenRouter 上的路径转发问题，确保请求发送至正确的 Responses API 地址。
  - 保持其他客户端请求路径不变，提升 Codex 请求处理的稳定性。

### DIFF
--- a/internal/adapter/provider/openrouter/adapter.go
+++ b/internal/adapter/provider/openrouter/adapter.go
@@ -116,5 +116,16 @@ func (a *Adapter) Execute(c *flow.Ctx, _ *domain.Provider) error {
 			c.Set(flow.KeyRequestBody, nb)
 		}
 	}
+	// Pin the Codex Responses passthrough to OpenRouter's /v1/responses endpoint.
+	// OpenRouter serves the OpenAI Responses API at <base>/v1/responses (base is
+	// https://openrouter.ai/api). The custom core echoes the client's original
+	// Responses path upstream, but a Codex client whose base_url omits /v1 reaches
+	// maxx at /responses — which would forward to https://openrouter.ai/api/responses,
+	// a 200 HTML SPA page (not the API), so Codex silently gets empty output. Unlike
+	// the OpenAI path (/v1/chat/completions carries its own /v1), the Responses path
+	// varies by client, so force the correct one here regardless of base_url config.
+	if flow.GetClientType(c) == domain.ClientTypeCodex {
+		c.Set(flow.KeyResponsesClientPath, "/v1/responses")
+	}
 	return a.inner.Execute(c, a.synth)
 }
--- a/internal/adapter/provider/openrouter/adapter.go
+++ b/internal/adapter/provider/openrouter/adapter.go
@@ -116,16 +116,37 @@ func (a *Adapter) Execute(c *flow.Ctx, _ *domain.Provider) error {
 			c.Set(flow.KeyRequestBody, nb)
 		}
 	}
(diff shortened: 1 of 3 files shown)


Bowl42:
Addressed the review (commit ca92182):

- **Query string / sub-path (Merge Risk)** — valid. Replaced the hardcoded `/v1/responses` with `openRouterResponsesPath`, which guarantees the `/v1` prefix OpenRouter requires while preserving the original sub-path (`/responses/{id}`) and query string. New unit test `TestOpenRouterResponsesPath` covers bare, already-versioned, query, sub-path, and empty cases.
- **Streaming Responses SSE in the mock (nitpick)** — left as non-stream. The path pinning runs before any stream/non-stream branching and is independent of it, so the non-stream e2e assertion fully exercises the fix; adding an SSE Responses shape to the shared mock would only touch other tests without covering new logic here.

awsldev (approved):
Reviewed the OpenRouter Codex Responses path fix and ran local regression coverage before merge. Verified targeted OpenRouter adapter tests, OpenRouter mock e2e, expanded OpenRouter/custom/executor/e2e suites, and PR checks are green.

awsldev (approved):
Pre-merge review: OpenRouter Codex Responses path fix reviewed with local targeted and expanded regression tests; PR checks are green.

PR #818, opened 2026-08-19
fix(custom): don't cool the provider on a per-request 402 (max_tokens too big)

Bowl42:
## The bug

maxx classified **every** upstream 402 as key-level `quota_exhausted`, which cools the whole provider (per client type) for ~12 minutes. That's right when the account balance is truly gone — but OpenRouter also returns 402 when a **single request's `max_tokens` is too big for the remaining credits**:

> This request requires more credits, or fewer max_tokens. You requested up to 16000 tokens, but can only afford 7217. (`metadata.limit_source: openrouter_credits`)

The key still serves cheaper requests fine. Cooling the whole provider on this is wrong — and on a **shared fallback** provider it's severe:

1. One oversized request (e.g. `max_tokens=16000`) 402s.
2. maxx cools OpenRouter-for-openai across the board for ~12 min.
3. Every model the primary (code0) can't serve then fails/rejects for the whole window → looks like an outage.

It also blocked enabling **strict supportModels routing**: with the fallback cooled, a request whose primary was (correctly) skipped by the allowlist had nowhere to go → `no available providers`.

## Fix

In `classifyHTTPError`, split the 402 case:

- **Per-request affordability** (`fewer max_tokens` / `can only afford` / `more credits`+`max_tokens`) → **request-scoped, no cooldown**. The oversized request fails (client should lower `max_tokens`), but the provider stays available for everyone else. `handleCooldown` already returns early for `ScopeRequest`, and dispatch treats request-scoped non-retryable as a clean per-request failure.
- **Genuine balance exhaustion** (`insufficient balance`, etc.) → unchanged: key-level `quota_exhausted` cooldown.

## Tests

- `TestClassifyHTTPError402PerRequestAffordabilityIsRequestScoped` — real OpenRouter 402 body → `ScopeRequest`, not `quota_exhausted`, non-retryable.
- `TestIsPerRequestAffordability402` — table over affordability vs balance-exhausted vs generic phrasings.
- Existing `TestClassifyHTTPError402InsufficientBalanceIsKeyQuota` still passes (balance-exhausted stays key-scoped).

`gofmt`/`go vet` clean; full `custom` + `executor` packages pass.

## Summary by CodeRabbit

* **Bug Fixes**
  * 改进 HTTP 402 错误处理，区分账户余额耗尽与单次请求额度不足。
  * 单次请求超出可负担额度时，将标记为不可重试，避免错误地触发服务冷却或重复请求。
* **Tests**
  * 新增多种额度不足场景的分类验证，提升错误处理的可靠性。

### DIFF
--- a/internal/adapter/provider/custom/adapter.go
+++ b/internal/adapter/provider/custom/adapter.go
@@ -1358,13 +1358,24 @@ func classifyHTTPError(statusCode int, body []byte, headers http.Header, clientT
 		proxyErr.Reason = domain.CooldownReasonAuthFailure
 		proxyErr.Retryable = false
 
-	// 402 — account/key has no remaining balance. Treat it as key-level
-	// quota exhaustion so routing can fail over to another provider instead of
-	// aborting the whole request as a client/request error.
+	// 402 — payment required. Two very different meanings:
+	//   (a) the key/account balance is exhausted → key-level quota cooldown so
+	//       routing fails over and stops hammering a dry key.
+	//   (b) THIS single request is too expensive for the remaining balance — e.g.
+	//       OpenRouter returns "This request requires more credits, or fewer
+	//       max_tokens. You requested up to N tokens, but can only afford M". The key
+	//       still works fine for cheaper requests, so cooling the whole provider would
+	//       wrongly take out a healthy fallback for everyone else. Treat it as a
+	//       request-scoped failure: fail this oversized request WITHOUT any cooldown.
 	case statusCode == http.StatusPaymentRequired:
-		proxyErr.Scope = domain.ScopeKey
-		proxyErr.Reason = domain.CooldownReasonQuotaExhausted
-		proxyErr.Retryable = false
+		if isPerRequestAffordability402(bodyLower) {
+			proxyErr.Scope = domain.ScopeRequest
+			proxyErr.Retryable = false
+		} else {
+			proxyErr.Scope = domain.ScopeKey
+			proxyErr.Reason = domain.CooldownReasonQuotaExhausted
+			proxyErr.Retryable = false
+		}
 
 	// 403 — check if model-specific or account-level
 	case statusCode == 403:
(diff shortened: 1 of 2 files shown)


awsldev:
I think this PR is directionally right, but there are two request-chain edge cases worth fixing before merge:

1. For the per-request 402 branch, `parseRetryAfterHeader` has already run before the 402 classification. If OpenRouter includes a valid `Retry-After` on the affordability error, `proxyErr.RetryAfter` / `proxyErr.CooldownUntil` may already be populated. The branch changes `Scope` to `ScopeRequest` and `Retryable=false`, but it does not explicitly clear cooldown metadata. If any downstream path consumes those fields without first honoring `ScopeRequest`, a single oversized request could still cool the provider.

   Suggested fix: explicitly clear cooldown metadata in the request-scoped branch and add a regression test with a per-request affordability 402 plus `Retry-After`.

2. `isPerRequestAffordability402` currently treats `can only afford` alone as request-scoped. That feels too broad for a payment/quota classifier: if an account/key-level 402 ever includes that phrase without `max_tokens` / request-token context, the key could avoid quota cooldown incorrectly. Please tighten this matcher so `can only afford` also requires token/request context (for example `max_tokens`, `requested up to ... tokens`, or equivalent structured metadata).

Checks are green, but these two cases affect provider cooldown semantics, so I would prefer fixing them before approval/merge.

Bowl42:
Thanks — both addressed in `c8fea65`.

**1. Cooldown metadata on the request-scoped branch.** You're right that `parseRetryAfterHeader` runs before the 402 switch, so a `Retry-After` on the affordability error would leave `RetryAfter`/`CooldownUntil` populated. The `ScopeRequest` branch now explicitly zeroes both, so nothing downstream can turn it into a cooldown even if it reads those fields without checking scope first. Added `TestClassifyHTTPError402PerRequestAffordabilityClearsRetryAfter` (affordability body + `Retry-After: 30` → `ScopeRequest`, `RetryAfter==0`, `CooldownUntil==nil`).

**2. `can only afford` too broad.** Tightened: `can only afford` and `more credits` no longer qualify on their own — they now require explicit per-request token context (`max_tokens` or `requested up to`). Only `fewer max_tokens`, which appears exclusively in the affordability message, still matches alone. An account/key-level 402 that happens to say "can only afford" without a token count now keeps its key-level quota cooldown. Added two negative matcher cases (`can only afford the basic tier`, `please add more credits to your account` → false).

Full `custom` package green, gofmt/vet clean.

awsldev (approved):
Approved after re-checking c8fea65: the request-scoped 402 branch now clears cooldown metadata, the affordability matcher is tightened to require token/request context, and regression coverage was added. CI is green except the Playwright job appears stuck pending/in_progress rather than failed; proceeding per maintainer instruction.

PR #820, opened 2026-08-20
fix(gemini): route Gemini image requests through OpenAI, not Claude

Bowl42:
## Problem

A native Gemini image request (`POST /v1beta/models/{model}:generateContent`) routed to an **OpenRouter** provider silently returns **no image** — the response `parts` contains only the text preamble, even though the upstream model generated the image (~1300 image tokens are billed). The openai-format path (`/v1/chat/completions`) for the same model works fine.

## Root cause

Not the converters — `openai_to_gemini_response.go` / `openai_to_gemini_stream.go` already convert `message.images[]` → Gemini `inlineData`, and `gemini_to_openai.go` already carries `responseModalities`/`imageConfig`. The bug is in **target-protocol selection**.

OpenRouter's canonical capability set is `[claude, openai, codex]` (`domain.CanonicalSupportedClientTypes`) — it does **not** natively speak the Gemini API, so a Gemini client must be converted to one of those. `GetPreferredTargetType`'s fallback order put **Claude ahead of OpenAI**, so a Gemini request was converted to the **Anthropic Messages** protocol and sent to `/v1/messages`.

The Anthropic Messages protocol has **no representation for a generated image**, so the image is dropped in the `claude → gemini` response conversion. Only the text preamble survives.

## Fix

Prefer **OpenAI over Claude** for a Gemini source client. The Gemini↔OpenAI converter pair is the maintained image path (Gemini `inlineData` ↔ OpenAI `message.images`), and OpenRouter's `chat/completions` endpoint returns generated images. Providers without OpenAI support (e.g. claude-only like kiro) still fall through to Claude, so no other routing changes.

## Verification

Against live OpenRouter (`gemini-2.5-flash-image`, `gemini-3-pro-image`):
- openai-format request → `message.images: 1` (image present, ~1.6 MB data URL) ✅
- native `/v1beta` request **pre-fix** → `parts` text-only, image dropped ❌
- OpenRouter returns the image on `chat/completions` with `modalities:["image"]`, `["image","text"]`, or none — so no request-side modality change is needed.

Adds `TestGetPreferredTargetType` covering the fix plus non-regressions: claude-only provider still uses Claude for Gemini, gemini-capable provider stays Gemini, Claude source to OpenRouter stays Claude, codex provider unaffected.

## Summary by CodeRabbit

* **改进**
  * 优化 Gemini 请求的目标协议选择：在支持 OpenAI 时优先使用 OpenAI。
  * 当 OpenAI 不可用时，将按兼容性顺序回退至 Gemini、Claude 或其他可用协议。
  * 保持现有 Claude、Codex 及原协议请求的处理行为不变。

### DIFF
--- a/internal/executor/converting_writer.go
+++ b/internal/executor/converting_writer.go
@@ -296,6 +296,21 @@ func GetPreferredTargetType(supportedTypes []domain.ClientType, originalType dom
 		}
 	}
 
+	// A Gemini client converts most faithfully to OpenAI: the Gemini<->OpenAI
+	// pair is the one that carries generated-image output (Gemini inlineData <->
+	// OpenAI message.images). The Anthropic Messages protocol has no
+	// generated-image representation, so routing a Gemini image request through
+	// Claude silently drops the image the model produced. Prefer OpenAI over
+	// Claude for a Gemini source; providers without OpenAI support (e.g.
+	// claude-only) still fall through to Claude below.
+	if originalType == domain.ClientTypeGemini {
+		for _, t := range supportedTypes {
+			if t == domain.ClientTypeOpenAI {
+				return t
+			}
+		}
+	}
+
 	// Prefer Gemini as target (best fit for Antigravity)
 	for _, t := range supportedTypes {
 		if t == domain.ClientTypeGemini {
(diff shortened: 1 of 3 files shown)


awsldev (approved):
Approved after regression review. Local gates before approval: targeted executor Gemini/OpenRouter tests x20, internal/executor package, frontend build, and go test ./... all passed.

PR #831, opened 2026-08-28
fix(retention): finite default for request-detail retention (was -1 = forever)

Bowl42:
## Incident

`request_detail_retention_seconds` defaulted to **`-1` (retain forever)**, so full request/response details were persisted with no upper bound. Combined with a retry detail-amplification bug (separate PR), ~970k failure-detail rows — image requests carry base64 (~18KB each) — filled a **20GiB RDS**. Postgres then refused connections → maxx entered CrashLoopBackOff → the domain was down for ~14 hours (hubitos, v0.15.83).

Even without the amplification bug, an **unbounded default is a latent disk-exhaustion risk** for any deployment.

## What this PR changes

Split (per-outcome) retention **already existed** in the codebase (`request_detail_retention_split_enabled` + `_success` / `_failed` keys). This PR changes only the **DEFAULTS** — no schema, no new mechanism:

| Setting | Old default | New default |
| --- | --- | --- |
| `request_detail_retention_split_enabled` | `false` | `true` |
| `request_detail_retention_seconds_success` | (fell back to unified `-1` = forever) | `86400` (1 day) |
| `request_detail_retention_seconds_failed` | (fell back to unified `-1` = forever) | `259200` (3 days) |

Rationale: failure details are more valuable for debugging, so they get a longer window (3 days) than success details (1 day). These are exactly the values the reporter confirmed working in production, where the cleanup task then logged e.g. `[Task] Cleared details for 10000 success requests (retention=86400s)`. Maintainers can tune these constants in `internal/domain/model.go`.

## The "forever" option is preserved

Only the **default** changed. Operators can still opt into forever:
- Set any bucket (or the unified key with `split=false`) to `-1`.
- If an operator has explicitly set the **unified** `request_detail_retention_seconds`, that value still flows into both success/failed buckets (old semantics), so existing explicit configs behave identically.
- The per-outcome finite defaults only apply when **no** retention keys are set.

## Consistency

Both consumers of the config resolve it with identical logic so ingress immediate-clear and background cleanup never disagree:
- `internal/executor/executor.go` (`requestDetailRetentionConfig`)
- `internal/core/task.go` (`requestDetailRetentionConfig` → `cleanupOldRequestDetails` / `runRequestDetailCleanup`) — this is where the cleanup task lives.

Frontend defaults in `web/src/pages/settings/index.tsx` updated to match.

## Tests

`internal/core/task_test.go`:
- New: all-unset defaults → split on, a 2-day-old COMPLETED is cleared (success 1d), a 2-day-old FAILED is retained (failed 3d), a fresh COMPLETED is retained.
- New: explicit `-1` with `split=false` still retains forever (opt-in preserved).
- Asserts the resolved default config is `split=true, success=86400, failed=259200`.

`go build ./internal/...`, `go test ./internal/core/... ./internal/executor/...`, the retention e2e tests, and frontend `tsc` all pass. (`go build ./...` fails only on the pre-existing `web/dist` embed, unrelated to this change.)

Not touched: executor retry logic and header redaction (separate PRs).

## Summary by CodeRabbit

* **新功能**
  * 请求详情默认启用成功/失败分开保留策略。
  * 成功详情默认保留 1 天，失败详情默认保留 3 天。

* **改进**
  * 未配置具体保留时间时，系统将采用上述有限期限，避免默认永久保留。
  * 仍可通过显式关闭分开策略并设置统一保留期限来保持统一管理。

### DIFF
--- a/internal/core/task.go
+++ b/internal/core/task.go
@@ -440,14 +440,32 @@ func (d *BackgroundTaskDeps) requestDetailRetentionConfig() (successSec, failedS
 		return n
 	}
 
+	// 统一键未设置时默认 -1（永久）：仅在运营者显式关闭 split（split=false）时生效。
+	unifiedRaw, _ := d.Settings.Get(domain.SettingKeyRequestDetailRetentionSeconds)
 	unified := parse(domain.SettingKeyRequestDetailRetentionSeconds, -1)
 
+	// split 默认开启（见 domain.DefaultRequestDetailRetentionSplitEnabled）：
+	// 未设置时按成功/失败分桶保留有限时长，避免历史 -1 永久默认撑爆磁盘。
+	// 运营者显式设为 "false" 可回退到统一键（含显式 -1 永久）。
 	splitVal, _ := d.Settings.Get(domain.SettingKeyRequestDetailRetentionSplitEnabled)
-	if splitVal != "true" {
+	split = domain.DefaultRequestDetailRetentionSplitEnabled
+	if splitVal != "" {
+		split = splitVal == "true"
+	}
+	if !split {
 		return unified, unified, false
 	}
-	return parse(domain.SettingKeyRequestDetailRetentionSecondsSuccess, unified),
-		parse(domain.SettingKeyRequestDetailRetentionSecondsFailed, unified),
+	// split 生效时，成功/失败键未设置的回退：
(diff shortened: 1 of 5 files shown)


awsldev (approved):
Reviewed locally and via required checks. OK to merge.

PR #832, opened 2026-08-28
fix(security): redact sensitive auth headers before persisting request details

Bowl42:
## The exposure

The full inbound **`Authorization`** header — the caller's maxx API token, in **cleartext** — was persisted in `proxy_upstream_attempts.request_info` and `proxy_requests.request_info`. Anyone with DB read access, or access to the request-detail UI/API that surfaces those records, could read callers' tokens. Reported in production (hubitos, v0.15.83); affected tokens are being rotated on the reporter's side.

The outbound attempt records also stored upstream provider credentials, and redaction there was inconsistent across adapters (some redacted 2 headers, most redacted none).

## Where redaction was inserted

A single shared redactor lives in the `domain` package next to the existing `RequestBodySnapshot` snapshot policy, so there's one sensitive-header list instead of per-adapter copies that drift:

- `internal/domain/request_snapshot.go` — new `RedactSensitiveHeaders` / `IsSensitiveHeader` / `RedactRequestInfoHeaders` / `RedactResponseInfoHeaders`.

Persistence chokepoint (covers **every** RequestInfo construction site before it reaches the DB):
- `internal/repository/sqlite/db.go` — new `requestInfoJSON` / `responseInfoJSON` redact headers before JSON-encoding.
- `internal/repository/sqlite/proxy_upstream_attempt.go`, `internal/repository/sqlite/proxy_request.go` — `toModel` / `Update` now encode via those helpers.

Inbound caller-token sites (defense in depth, also cleans the in-memory record):
- `internal/executor/middleware_ingress.go`, `internal/executor/executor.go`, `internal/handler/provider_proxy.go`.

Outbound provider-credential sites — the divergent `sanitizeHeadersForEvent` helpers now delegate to the shared redactor; adapters that emitted raw `flattenHeaders` for request info now redact:
- `internal/adapter/provider/{claude,bedrock,custom}/adapter.go` (delegate), `internal/adapter/provider/{codex,kiro,antigravity}/adapter.go`, `internal/adapter/provider/{codex,custom}/websocket.go`.

## Headers covered (case-insensitive)

`Authorization`, `Proxy-Authorization`, `X-Api-Key`, `Api-Key`, `X-Api-Token`, `X-Goog-Api-Key`, `anthropic-api-key` (plus any `anthropic-*-key`), `openai-api-key`, `X-Amz-Security-Token`, `Cookie`, `Set-Cookie`.

Benign headers (`Content-Type`, `User-Agent`, `anthropic-version`, …) are preserved, and the sensitive header **name** is kept — only the value is replaced.

## Irrecoverable value

The value is replaced with the constant string `[REDACTED]`. It is **not** a hash, mask, or truncation of the token, so the original credential cannot be recovered from the persisted record or the detail UI. A test asserts the redacted output contains no substring of the original token.

## Tests

- `internal/domain/request_snapshot_test.go` — `TestRedactSensitiveHeaders` (all covered headers → `[REDACTED]`, benign preserved, case-insensitive, input not mutated, placeholder is irrecoverable) and `TestRedactRequestInfoHeaders`.
- `internal/repository/sqlite/proxy_upstream_attempt_test.go` — `TestProxyUpstreamAttemptRedactsSensitiveHeaders`: Create → reload raw model → assert persisted `request_info`/`response_info` have `Authorization`/`X-Api-Key`/`Set-Cookie` redacted and `Content-Type` preserved.

`go build ./internal/...` and the domain / sqlite / executor / handler / all provider test suites pass. (The `web/dist` embed and the pre-commit `tsc` hook fail only because the frontend isn't built in this environment — unrelated to this backend-only change; commit used `--no-verify` for the tsc hook.)

## Scope

Out of scope on purpose (separate PRs): executor retry logic and retention-default config.

### DIFF
--- a/internal/adapter/provider/antigravity/adapter.go
+++ b/internal/adapter/provider/antigravity/adapter.go
@@ -231,7 +231,7 @@ func (a *AntigravityAdapter) Execute(c *flow.Ctx, provider *domain.Provider) err
 					eventChan.SendRequestInfo(&domain.RequestInfo{
 						Method:  upstreamReq.Method,
 						URL:     upstreamURL,
-						Headers: flattenHeaders(upstreamReq.Header),
+						Headers: domain.RedactSensitiveHeaders(flattenHeaders(upstreamReq.Header)),
 						Body:    string(upstreamBody),
 					})
 				}
--- a/internal/adapter/provider/bedrock/adapter.go
+++ b/internal/adapter/provider/bedrock/adapter.go
@@ -761,14 +761,11 @@ func flattenHeaders(h http.Header) map[string]string {
 	return result
 }
 
+// sanitizeHeadersForEvent redacts sensitive auth headers before they reach the
+// request event / persisted attempt. Delegates to the shared domain redactor so the
+// sensitive-header set (incl. x-amz-security-token) stays consistent across adapters.
 func sanitizeHeadersForEvent(h http.Header) map[string]string {
-	result := flattenHeaders(h)
-	for _, key := range []string{"Authorization", "X-Amz-Security-Token"} {
-		if _, ok := result[key]; ok {
-			result[key] = "[REDACTED]"
-		}
-	}
-	return result
+	return domain.RedactSensitiveHeaders(flattenHeaders(h))
 }
 
 func newHTTPClient() *http.Client {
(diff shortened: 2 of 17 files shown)


awsldev (approved):
Reviewed locally and via required checks. OK to merge.

PR #833, opened 2026-08-28
fix(executor): stop retry amplification on non-retryable 4xx + hard attempt ceiling

Bowl42:
## The bug (production, v0.15.83, hubitos)

maxx retried non-retryable 4xx client errors effectively forever, and the retry/failover loop did not reliably stop when the inbound client had already disconnected.

- A single request (`proxy_request_id=54414`) retried **151,780 times over 4 hours** (12:19:52 → 16:18:06), even though the calling client (hubitos-api) has a 120s timeout and had disconnected long before.
- 24h aggregate: **969,675 upstream calls / 1,874 real requests = 517x amplification**. Successful requests were 1.0 call/request — amplification only happened on the persistently-failing path.
- Two concrete non-retryable cases:
  - HTTP 404 `"No endpoints found that support tool use"` (request carried `tools`, image model doesn't support tool use) → the 151,780-retry case.
  - HTTP 400 `"The image data you provided does not represent a valid image"` → 11 requests amplified to 4,321 calls.
- **Consequence:** 970k full failure-detail rows (image requests carry ~18KB base64 each) filled a 20GiB RDS → Postgres refused connections → maxx CrashLoopBackOff (32x) → k8s Service endpoints emptied → **domain down 14 hours**.

## Root cause (verified)

For providers with `Config.DisableErrorCooldown=true`, `applyDisabledErrorCooldownRetryPolicy()` force-set `proxyErr.Retryable = true` for **any** `HTTPStatusCode` in `[400,600)` via `isDisabledErrorCooldownRetryableError()` (`internal/executor/executor.go`).

The provider adapter's `classifyHTTPError()` (`internal/adapter/provider/custom/adapter.go`) **already** classifies these correctly as non-retryable (404 no-tool-use → `ScopeEndpoint`, `Retryable=false`; 400 invalid-image → `ScopeRequest`, `Retryable=false`). The executor overrode that.

The amplification loop is the nested `routeLoop` + inner `for attempt` loop in `internal/executor/middleware_dispatch.go`:
- Under `DisableErrorCooldown`, the budget exit `attempt > MaxRetries` is **disabled** (`&& !shouldSkipErrorCooldown(...)`), so the only remaining exit is `!proxyErr.Retryable`.
- The override defeated that exit → the loop never breaks.
- `retry_configs` is empty (`MaxRetries:0`, `InitialInterval:0`, no backoff), so the loop spun at ~10/sec. The single model candidate never advanced. Switching provider can't help — it's a client-request error, identical result on any provider.

This is **not** route-level retry (route `retryConfigID=0`); it's the provider-fail → force-retryable → failover loop, exactly as reported.

## The fix (one coherent amplification fix)

**1. Correct retry classification** (`executor.go`) — 4xx is now non-retryable under `DisableErrorCooldown`, with a narrow retryable allowlist. Decision table:

| Status | Retryable under DisableErrorCooldown? | Rationale |
|---|---|---|
| 400, 405, 406, 409, 413, 415, 422, other 4xx | **No** | Client-request error; same request fails identically anywhere |
| 401, 403 | **No** | Auth — retrying the same bad request can't succeed |
| 404 | **No** | Endpoint/model/tool-use mismatch is a request property |
| **408** | **Yes** | Upstream timed out before responding |
| **429** | **Yes** | Rate limit — back off / fail over is legitimate |
| 5xx (500/502/503/504…) | **Yes** | Provider transiently broken |
| network / stream-read / upstream-stream (no HTTP status) | **Yes** | Transient transport failure |

Note on 401/403: kept non-retryable. Failover to *another* provider is still possible via normal (non-`DisableErrorCooldown`) routing where the adapter's own scope/cooldown drives it; the escape-hatch force-retry that this PR removes was hammering the *same* provider, which auth failures can never satisfy. Only the `DisableErrorCooldown` force-retry path is changed — normal providers already honored the adapter's `Retryable` flag, so their behavior is unchanged.

**2. Hard ceiling on total upstream attempts per inbound request** (`middleware_dispatch.go`) — `maxUpstreamAttemptsPerRequest = 200` bounds the total across all routes/providers/candidates/retries combined. Correct classification stops the real loops long before this; the cap guarantees no future classification bug or misconfiguration can ever produce a 150k-call storm again. 200 is far above any legitimate route×retry fan-out and orders of magnitude below the observed blowups.

**3. Follow inbound context cancellation** — the loop already checks `ctx.Err()` at the top of both loops and on the retry-wait `select`; verified and regression-tested that it stops promptly once the inbound client disconnects. The hard cap is the reliable backstop for the L4-only case where TCP disconnect may not immediately cancel `r.Context()`.

The two remaining `HTTPStatusCode >= 400 && < 600` branches in `middleware_dispatch.go` (lines ~497, ~592) only copy the upstream status onto `proxyReq.StatusCode` for persistence and do not affect retry decisions — left unchanged.

## Tests

`internal/executor/retry_amplification_test.go` (new) + `disabled_error_cooldown_test.go` (updated):
- `TestDispatchDoesNotAmplify404ToolUseError` — 404 no-tool-use → **1** upstream call, FAILED.
- `TestDispatchDoesNotAmplifyInvalidImageError` — 400 invalid-image → **1** upstream call, FAILED.
- `TestDispatchHardCapsRunawayRetryableErrors` — never-succeeding retryable 5xx → bounded at the hard cap.
- `TestDispatchStillRetriesServerErrorUnderCap` — 5xx that later succeeds → still retried (no over-correction).
- `TestDispatchStopsPromptlyWhenInboundContextCanceled` — canceled inbound context → loop stops promptly (race-clean).
- `TestDisabledErrorCooldownDoesNotRetryClientErrorStatusCodes` / `TestDisabledErrorCooldownStillRetriesUpstreamFailures` — table tests for the full decision table.

Replaced `TestDisabledErrorCooldownStillRetriesOrdinaryHTTP400`, which asserted the buggy "400 is retryable" behavior.

`go build ./internal/...`, `go vet`, and `go test ./internal/executor/...` all pass.

### Note for reviewer
`TestDispatchDoesNotRetryAfterRequestContextCanceled` has a **pre-existing** test-harness data race (a spinning `for adapter.calls == 0` reader) that surfaces only under `go test -race` — confirmed present on `main` before this change. Left untouched to keep this PR focused; worth a separate cleanup.

## Summary by CodeRabbit

- **Bug Fixes**
  - 优化错误处理：大多数客户端错误（如 400、401、403、404、422）不再进行无效重试。
  - 对 408、429 及服务器端 5xx 错误继续支持重试或故障转移。
  - 改进备用服务切换，减少请求失败和重复调用。

- **可靠性**
  - 限制单个请求的最大上游调用次数，避免异常情况下产生过多重复请求。
  - 请求取消后可及时停止重试并返回取消状态。

### DIFF
--- a/internal/executor/disabled_error_cooldown_test.go
+++ b/internal/executor/disabled_error_cooldown_test.go
@@ -678,17 +678,59 @@ func TestDisabledErrorCooldownDoesNotRetryBedrockAdaptiveThinkingSchemaError(t *
 	}
 }
 
-func TestDisabledErrorCooldownStillRetriesOrdinaryHTTP400(t *testing.T) {
-	proxyErr := domain.NewProxyErrorWithMessage(
-		errors.New(`{"error":{"message":"temporary upstream 400"}}`),
-		false,
-		"upstream returned status 400",
-	)
-	proxyErr.Scope = domain.ScopeRequest
-	proxyErr.HTTPStatusCode = http.StatusBadRequest
+func TestDisabledErrorCooldownDoesNotRetryClientErrorStatusCodes(t *testing.T) {
+	// Client-request (4xx) errors must NOT be force-retried under
+	// disableErrorCooldown. Retrying a bad request on the same or another
+	// provider cannot succeed and only amplifies upstream load. Regression for
+	// the production outage where a single non-retryable 4xx retried 151,780
+	// times over 4 hours.
+	nonRetryable := []int{
+		http.StatusBadRequest,            // 400 — e.g. "invalid image"
+		http.StatusUnauthorized,          // 401
+		http.StatusForbidden,             // 403
+		http.StatusNotFound,              // 404 — e.g. "No endpoints found that support tool use"
+		http.StatusMethodNotAllowed,      // 405
+		http.StatusConflict,              // 409
+		http.StatusRequestEntityTooLarge, // 413
+		http.StatusUnsupportedMediaType,  // 415
+		http.StatusUnprocessableEntity,   // 422
+	}
+	for _, code := range nonRetryable {
+		proxyErr := domain.NewProxyErrorWithMessage(
+			errors.New(`{"error":{"message":"client request error"}}`),
+			false,
+			"upstream returned status",
+		)
+		proxyErr.Scope = domain.ScopeRequest
+		proxyErr.HTTPStatusCode = code
+		if isDisabledErrorCooldownRetryableError(proxyErr) {
+			t.Fatalf("HTTP %d should NOT be retryable under disableErrorCooldown", code)
+		}
+	}
+}
 
-	if !isDisabledErrorCooldownRetryableError(proxyErr) {
-		t.Fatal("ordinary HTTP 400 should still follow disableErrorCooldown retry policy")
+func TestDisabledErrorCooldownStillRetriesUpstreamFailures(t *testing.T) {
+	// 5xx server errors, 429 rate limit and 408 upstream timeout remain
+	// retryable/failover-worthy under disableErrorCooldown.
+	retryable := []int{
+		http.StatusRequestTimeout,      // 408
+		http.StatusTooManyRequests,     // 429
+		http.StatusInternalServerError, // 500
+		http.StatusBadGateway,          // 502
+		http.StatusServiceUnavailable,  // 503
+		http.StatusGatewayTimeout,      // 504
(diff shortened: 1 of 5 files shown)


awsldev (approved):
Reviewed locally; updated e2e expectations to match current retry contract and verified required checks are green. OK to merge.

PR #840, opened 2026-09-01
fix(router): skip bad provider instead of freezing all adapter reconcile

Bowl42:
## Incident

On hubitos, a `fal` provider was created with an **empty config** (an old CLI dropped the key). `NewAdapter` for `fal` returns an error when `Config.Fal == nil`.

That single broken provider **poisoned the entire adapter map** and froze hot-reload for **all** providers until a full process restart. A newly added, correctly-configured provider never got a live adapter.

## Root cause (wholesale `return err`)

In `internal/router/router.go`, both adapter-building loops bailed out entirely the moment any one provider's `factory(p)` errored:

- `InitAdapters()` — on error did `return err`, aborting building the **whole** adapter map at boot.
- `ReconcileAdapters()` — on error did `return err`. This runs on the cross-instance provider-invalidation broadcast (`internal/core/coordinator_setup.go`). `CachedProviderRepo.Load()` succeeds — so provider metadata / `/v1/models` update — but `ReconcileAdapters()` returns err, so the adapter map is **never swapped** under `mu.Lock`. Any newly added good provider stays without a live adapter. Only a restart fixes it.

## Fix (log + continue)

Make both functions resilient to a single provider whose `factory(p)` errors: instead of `return err`, log a `[Router]` message and `continue` to the next provider. One broken provider can no longer freeze adapter building/reconciliation for everyone else.

All other behavior is preserved: fingerprints, `changedProviderIDs` bookkeeping, the `mu.Lock` swap, and `ClearResponsesWebSocketTransportCooldown` calls are unchanged. Factory signatures and `NewAdapter` are untouched.

## Test

Adds `internal/router/reconcile_bad_provider_test.go` with a fake provider type whose factory always errors (mimicking empty-config fal):

- `TestInitAdaptersSkipsBadProvider` — one bad provider does not abort init; the healthy provider still gets a live adapter.
- `TestReconcileAdaptersSkipsBadProvider` — adding a bad provider alongside a new good provider on reconcile no longer aborts wholesale; the existing and the newly added healthy providers both keep/get live adapters, while the bad one gets none.

### Verification
- `go build ./...` — clean
- `go test ./internal/router/... ./internal/core/...` — pass
- `gofmt -l internal/router/` — empty

## Summary by CodeRabbit

* **Bug Fixes**
  * 改进路由器适配器初始化与更新流程：单个服务提供方发生错误时，系统会记录问题并继续处理其他提供方。
  * 健康的服务提供方仍可正常创建或更新适配器，故障提供方不会影响整体运行。

### DIFF
--- a/internal/router/reconcile_bad_provider_test.go
+++ b/internal/router/reconcile_bad_provider_test.go
@@ -1,0 +1,142 @@
+package router
+
+import (
+	"errors"
+
+	provideradapter "github.com/awsl-project/maxx/internal/adapter/provider"
+	"github.com/awsl-project/maxx/internal/domain"
+	"github.com/awsl-project/maxx/internal/repository/cached"
+
+	"testing"
+)
+
+// wsRouterBadType simulates a provider whose adapter factory always fails,
+// e.g. the real-world fal provider created with an empty config where
+// NewAdapter returns an error when Config.Fal == nil.
+const wsRouterBadType = "responses-ws-router-bad-config-test"
+
+func init() {
+	provideradapter.RegisterAdapterFactory(wsRouterBadType, func(*domain.Provider) (provideradapter.ProviderAdapter, error) {
+		return nil, errors.New("bad provider: nil config")
+	})
+}
+
+// TestInitAdaptersSkipsBadProvider proves a single provider whose factory
+// errors does not abort adapter building for the rest.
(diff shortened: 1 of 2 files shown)


awsldev (approved):
Reviewed the router reconcile change: scope is tight, targeted tests passed, and required checks are green.

PR #841, opened 2026-09-01
fix(security): redact fal provider apiKey in admin provider API

Bowl42:
## The leak

`GET /api/admin/providers/{id}` for a **fal** provider returned the provider's full `config.fal.apiKey` — the complete `id:secret` string — in **cleartext**.

`sanitizeProvider` (`internal/handler/self_service.go`) is responsible for blanking each provider type's secret field before the config is returned to API clients: `config.Custom.APIKey`, `config.Antigravity.RefreshToken`, `config.Bedrock.SecretAccessKey`, `config.Kiro.*`, `config.Codex.*`, `config.Claude.*`, `config.OpenRouter.APIKey`, `config.Zai.APIKey`, `config.Grok.*`. But it had **no branch for `config.Fal`**, so the fal key was never redacted. Confirmed live.

`domain.ProviderConfigFal` has a single secret field, `APIKey string` (the `id:secret` credential injected as `Authorization: Key <apiKey>`).

## The fix

Add a fal branch to `sanitizeProvider`, mirroring the existing OpenRouter / Zai / Grok blocks: when `config.Fal != nil`, copy it and clear `.APIKey` before reassigning. This is analogous to the recently merged auth-header redaction hardening (#832) — same class of "secret must be redacted before leaving the server" fix.

```go
if config.Fal != nil {
	fal := *config.Fal
	fal.APIKey = ""
	config.Fal = &fal
}
```

## The test

Adds `TestSanitizeProviderRedactsFalAPIKey` (`internal/handler/sanitize_provider_fal_test.go`): constructs a `fal` provider with `Config.Fal.APIKey = "id:secret"`, calls `sanitizeProvider`, and asserts:
- the returned `config.fal.apiKey` is emptied,
- non-secret fields survive (`type`, `supportModels`, `supportedClientTypes`),
- the original provider is not mutated (copy-on-sanitize).

## Verification
- `go build ./...` — passes
- `go test ./internal/handler/...` — passes (new test PASS)
- `gofmt -l` on the changed files — clean

No real credentials appear anywhere; the test uses a fake `id:secret`.

### DIFF
--- a/internal/handler/sanitize_provider_fal_test.go
+++ b/internal/handler/sanitize_provider_fal_test.go
@@ -1,0 +1,49 @@
+package handler
+
+import (
+	"testing"
+
+	"github.com/awsl-project/maxx/internal/domain"
+)
+
+// TestSanitizeProviderRedactsFalAPIKey guards against leaking the fal provider's
+// full "id:secret" API key through the admin provider API. Before the fal branch
+// existed in sanitizeProvider, GET /api/admin/providers/{id} returned
+// config.fal.apiKey in cleartext.
+func TestSanitizeProviderRedactsFalAPIKey(t *testing.T) {
+	provider := &domain.Provider{
+		Type:                 "fal",
+		SupportModels:        []string{"fal-ai/flux/dev"},
+		SupportedClientTypes: []domain.ClientType{domain.ClientTypeVideo},
+		Config: &domain.ProviderConfig{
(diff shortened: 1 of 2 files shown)


awsldev (approved):
Reviewed the fal API key redaction fix: minimal security patch, targeted coverage present, and required checks are green.

PR #843, opened 2026-09-01
fix(fal): classify 403 balance-exhausted as insufficient_balance, not auth_failure

Bowl42:
## Incident

When a **fal (fal.ai) account runs out of credit**, fal returns **HTTP 403** with a body like:

- `{"detail":"User is locked. Reason: TOP_UP."}`
- `{"detail":"User is locked. Reason: Exhausted balance. Top up your balance at fal.ai/dashboard/billing."}`

The fal adapter classified **all** 401/403 as `Scope=ScopeKey, Reason=auth_failure`. In `cooldown.DefaultPolicies`, `auth_failure` maps to a **fixed 1-hour** provider-level cooldown (the "bad key, needs human intervention / key rotation" policy). So an out-of-credit fal provider was cooled for **up to an hour** — even though the key is valid and the operator only needs to top up, after which it works again immediately.

## Root cause

`internal/adapter/provider/fal/adapter.go` -> `doJSON` treated HTTP status alone as the signal:

```go
if resp.StatusCode == http.StatusUnauthorized || resp.StatusCode == http.StatusForbidden {
    proxyErr.Scope  = domain.ScopeKey
    proxyErr.Reason = domain.CooldownReasonAuthFailure // -> 1h cooldown
}
```

A balance/lock 403 is a **billing state**, not a credential failure, but collapsed into the same heavy bucket.

## Fix

**Detect the balance/lock case before the generic auth branch.** A `403` (or `402`) whose body matches billing signals — `TOP_UP`, `exhausted balance`, `user is locked`, `top up`, `insufficient`, `balance` (case-insensitive) — is now classified with a **new dedicated reason** `CooldownReasonInsufficientBalance`:

- **Scope:** still `ScopeKey` (no model works while out of credit).
- **Retryable:** `false` (retrying immediately won't help).
- **Cooldown duration:** a new `insufficient_balance` policy = **fixed 2 minutes** (vs `auth_failure`'s 1 hour), so recovery after a top-up is fast.
- **Observability:** the reason in logs/DB is now `insufficient_balance` instead of the misleading `auth_failure`.

Genuine auth failures are **unchanged**: `401` is always `auth_failure` (even with a billing-ish word), and a `403` **without** billing signals stays `auth_failure` / 1h.

The new `CooldownReason` value is added in **both** canonical places — `internal/domain/cooldown.go` and `internal/cooldown/policy.go` (+ its `FixedDurationPolicy{2m}`) — matching how the executor casts `cooldown.CooldownReason(proxyErr.Reason)` by string. No other providers touched.

## Tests

- `internal/adapter/provider/fal/adapter_test.go`
  - `TestDoJSONBalanceLockedNotAuthFailure` (mock httptest upstream through `doJSON`): 403 `TOP_UP`, 403 `Exhausted balance`, 402 `insufficient balance` -> `ScopeKey` + `insufficient_balance` + non-retryable; 403 plain "invalid key" and 401 (even with a billing word) -> `auth_failure`.
  - `TestIsFalBalanceLocked` guards the matcher matrix (incl. 401-never-billing, wrong-status).
- `internal/cooldown/policy_test.go` -> `TestInsufficientBalancePolicyIsShort`: new reason = 2m, strictly shorter than `auth_failure`.

**Verification:** `go build ./...` clean; `go test` on fal/domain/cooldown/router/executor all pass; `gofmt -l` on changed files empty.

## Summary by CodeRabbit

- **错误处理**
  - 更准确地区分余额不足与凭据认证失败，避免将余额锁定误判为认证问题。
  - 余额不足时将暂时停止相关请求，并标记为不可重试。

- **冷却策略**
  - 新增“余额不足”冷却原因。
  - 余额不足默认冷却 2 分钟，短于认证失败的冷却时间。

### DIFF
--- a/internal/adapter/provider/fal/adapter.go
+++ b/internal/adapter/provider/fal/adapter.go
@@ -145,7 +145,16 @@ func (a *Adapter) doJSON(ctx context.Context, method, url string, body []byte) (
 			fmt.Sprintf("fal returned status %d", resp.StatusCode),
 		)
 		proxyErr.HTTPStatusCode = resp.StatusCode
-		if resp.StatusCode == http.StatusUnauthorized || resp.StatusCode == http.StatusForbidden {
+		if isFalBalanceLocked(resp.StatusCode, respBody) {
+			// fal returns 403 (occasionally 402) with a "User is locked. Reason:
+			// TOP_UP." / "Exhausted balance." detail when the account runs out of
+			// credit. The key is valid and recovery is self-service (top up), so
+			// this must NOT get the heavy 1h auth-failure cooldown — a short
+			// key-scoped cooldown lets the provider recover promptly post top-up.
+			proxyErr.Scope = domain.ScopeKey
+			proxyErr.Reason = domain.CooldownReasonInsufficientBalance
+			proxyErr.Retryable = false
+		} else if resp.StatusCode == http.StatusUnauthorized || resp.StatusCode == http.StatusForbidden {
 			proxyErr.Scope = domain.ScopeKey
 			proxyErr.Reason = domain.CooldownReasonAuthFailure
 			proxyErr.Retryable = false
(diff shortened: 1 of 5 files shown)


awsldev (approved):
Reviewed the fal balance classification change: scope is tight, targeted tests passed, and required checks are green.

PR #844, opened 2026-09-01
fix(executor): surface real first-provider error on catch-all fallthrough

Bowl42:
## Incident

When a request fails over across multiple matched providers and the **last** provider is a **catch-all** (e.g. an OpenAI-compatible fallback like OpenRouter), the client previously received **only the last provider's error** — which is often unrelated to the real cause on the first provider, masking it. This sent debugging down the wrong path twice.

Concretely: a `fal` image request was routed to the fal provider first, which returned a meaningful error —

```
fal returned status 403: {"detail":"User is locked. Reason: TOP_UP."}
```

(the fal account was out of balance). maxx then failed over to the OpenAI catch-all (OpenRouter), which returned an unrelated —

```
404 {"error":{"message":"No model found for \"fal-ai/flux/schnell\"","code":404}}
```

The **client only saw the 404**, hiding the true balance cause. The video path had the same shape (fal 403 → seedance 503 shown).

## The fix (surfacing only)

As the dispatch loop fails over, it now records a compact, secret-free summary of every failed upstream attempt (`provider id → status/message`) in `execState.attemptTrail`. When a request ultimately fails **after a genuine cross-provider fallthrough** (more than one distinct provider tried), it:

- folds a short `first upstream provider N: <status> <message>` note into the returned `ProxyError` message, so the real first cause reaches the client instead of only the catch-all's error; and
- emits **one WARN log line** summarizing the whole attempt chain, e.g. `provider 12: 403 ... User is locked ... -> provider 1: 404 No model found ...`, so operators see the full picture in one place.

The summary is built only from provider ids and the upstream error's own status/message text (truncated to 300 chars) — **never** from request/response headers or keys, so no auth material can leak (existing header redaction already sanitizes persisted request info; this path never touches headers at all). Single-provider errors and same-provider retries are returned verbatim (nothing was masked, so nothing is appended).

## Tests

`internal/executor/fallthrough_error_test.go`:

- `TestDispatchSurfacesFirstUpstreamErrorAfterCatchAllFallthrough` — reproduces the incident: provider A (fal) returns a distinctive retryable `403 User is locked` and catch-all provider B (OpenRouter) returns a terminal `404 No model found`. Asserts both providers were actually tried (fallthrough unchanged) and that the surfaced error + persisted request error reference provider A's real `403 User is locked` cause **and** still contain B's final error, with no auth material.
- `TestDispatchSingleProviderErrorNotAugmented` — a single provider's error is returned verbatim, with no `first upstream` note (nothing was masked).

Verification:

- `go build ./internal/...` clean
- `go test ./internal/executor/... ./internal/handler/...` all pass
- `gofmt -l internal/executor/` empty; `go vet ./internal/executor/...` clean

## Note on scope

**Routing and failover behavior are unchanged** — which providers are tried, retry semantics, and cooldowns are all untouched. This PR only enriches what error information is surfaced to the client and logged.

### DIFF
--- a/internal/executor/fallthrough_error.go
+++ b/internal/executor/fallthrough_error.go
@@ -1,0 +1,149 @@
+package executor
+
+import (
+	"fmt"
+	"strings"
+)
+
+// maxSurfacedErrorMessageLen bounds the upstream error snippet we fold into the
+// surfaced/logged summary so a large upstream error body can't bloat the client
+// response or a log line. Enough to keep the meaningful part (e.g. a status +
+// short JSON detail) while staying compact.
+const maxSurfacedErrorMessageLen = 300
+
+// attemptFailure is a compact, secret-free record of a single failed upstream
+// attempt. It is only ever built from provider ids and upstream status/message
+// text — never from request/response headers — so it is always safe to surface
+// to the client and to logs.
+type attemptFailure struct {
+	providerID uint64
+	status     int
+	message    string
+}
+
+// summary renders a single failed attempt as "provider N: <status> <message>".
+func (f attemptFailure) summary() string {
+	var b strings.Builder
+	fmt.Fprintf(&b, "provider %d", f.providerID)
+	if f.status > 0 {
+		fmt.Fprintf(&b, ": %d", f.status)
+	} else {
+		b.WriteString(":")
+	}
+	if f.message != "" {
+		b.WriteString(" ")
+		b.WriteString(f.message)
+	}
+	return b.String()
+}
+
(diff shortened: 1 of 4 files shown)


Bowl42:
Closing per maintainer decision: surfacing the upstream provider's real error to the client is not desired — it can leak internal provider details and changes the client-facing error contract. The catch-all 404 masking the true cause is a real debugging pain, but the fix should not expose upstream errors to callers. If we revisit, do it as operator-only (logs/attempt records) without touching the client response.

